PR #167, opened 2024-01-02
Add Chinese translation of readme

502y:
And add a link to translation in README.md .
Also, if pages in [website](repentogon.com/install.html) can be translated?


wofsauge:
Thanks for the translation of the readme.

For the question of translating the website, that is possible to accomplish by using the following feature of MkDocs: https://squidfunk.github.io/mkdocs-material/setup/changing-the-language/

it basically requires us to change the docs folder structure a bit by adding language specific subfolders. It also requires us to keep a duplicate of each file, one per language.

Are there any specific files you would like to translate, or do you want to translate all the sites, including function documentations?

502y:
> Thanks for the translation of the readme.
> 
> For the question of translating the website, that is possible to accomplish by using the following feature of MkDocs: https://squidfunk.github.io/mkdocs-material/setup/changing-the-language/
> 
> it basically requires us to change the docs folder structure a bit by adding language specific subfolders. It also requires us to keep a duplicate of each file, one per language.
> 
> Are there any specific files you would like to translate, or do you want to translate all the sites, including function documentations?

Thank you very much! I am currently translating the text in #217 and I think I will start translating the website after the translation is complete.

PR #194, opened 2024-01-03
ColorParams

namishere:
This class defines the set of Colors stored by Entities via SetColor, alongside their relevant parameters (Priority, Duration, Fadeout, Shared). This allows these parameters to be edited, or copied over to another Entity to synchronize them.


namishere:
This also allows Entity:CopyStatusEffects() to copy the status effect color to the other entity.

namishere:
After testing the constructor, its functions, Entity:Get/SetColorParams, as well as the new CopyStatusEffects, I believe this is good to go.

PR #215, opened 2024-01-04
TrySpawnSpecialQuestDoor( boolean IgnoreStageType )

namishere:
This pull request adds an `IgnoreStageType` parameter to the vanilla function `TrySpawnSpecialQuestDoor`. This is in service of mods currently attempting to mimic alt-path floors without actually being on the alt-path StageTypes.


namishere:
After testing, door still both doesn't naturally spawn on non alt-path floors and does spawn on alt-path floors, and using the new argument successfully spawns the door, so...

PR #216, opened 2024-01-04
1.0.2 injector (for CI mostly)

Sylmir:


namishere:
This should be closable now, right?

PR #271, opened 2024-01-17
Minimap Update

epfly6:


libzhl/functions/MinimapConfig.zhl
```diff
@@ -0,0 +1,13 @@
+struct MinimapConfig depends (ANM2, Vector) {
+	ANM2 _anm2 : 0x0;
```
wofsauge: What are the member variables used for? We should avoid adding non-descriptive variable names and function calls where ever possible.

epfly6: Unknown, yet. When it is clear what the variable is responsible for, it will be renamed before merge

PR #296, opened 2024-01-27
Docs: Game():GetPauseMenuState returns a PauseMenuStates

Gemboat:


jsgnextortex:
This doesnt belong on Game, you can expect this on the next version alongside a lot of PauseMenu stuff we seemed to miss on the new PauseMenu global (which works like all the other menus we already have for consistency)

jsgnextortex:
I misunderstood this one....oh well, you can still expect what I said (although some redundancy will exist with this function that should have never been in Game to begin with)

PR #396, opened 2024-04-02
RoomConfig additions and refactoring

namishere:
This does a few things:

1. Updates ZHL definitions of `RoomConfig` and related classes to match current knowledge. It is now much closer to how the class is set up ingame.
2. Adds new lua classes `RoomConfigStage` and `RoomSet`, which allow direct access and modification of the parameters defined in `stages.xml`.
3. Add gfx suffixes (as used in Flooded Caves, Downpour, Dross, Ashpit, Mausoleum, Gehenna)
3. Clean up a bunch of compile warnings and a section of `ConsoleMega.h`.

A couple considerations:
* `RoomConfigStage:GetRoomSet` will provide `RoomSet` userdata even if none exists for that `RoomConfigStage`. Validity can be checked with `RoomSet:IsLoaded`.
* There's a non-zero chance that these changes could mess with console autocomplete with `stage` and `goto`, but I haven't had any come up yet.
* No documentation pages right now, I want to save that for near the release of next version.


libzhl/functions/Music.zhl
```diff
@@ -46,6 +46,16 @@ struct Music {
 		*(int*)((char*)this + 0x348) = 0;
 		*(int*)((char*)this + 0x364) = -99;
 	};
+	
+	inline bool ValidateMusicID(int id, int& max) {
```
Sylmir: You should declare a dummy class in ZHL (MusicInstance ?) with a fixed size of 96, and add two pointers to said class in the Music class so you can use types and offsets instead of translating the ASM into C

libzhl/functions/RoomConfigRoom.zhl
```diff
@@ -0,0 +1,28 @@
...
+    int8_t Width : 0x46; // 0x46
+    int8_t Height : 0x47; // 0x47
+    int32_t Shape : 0x48; // 0x48
+    int32_t unkFlags : 0x4c; // 0x4C
+    int32_t unk10 : 0x50; // 0x50
+    int32_t unk11 : 0x54; // 0x54
+    int32_t originalVariant : 0x58; // 0x58
+
+    {{
+    bool LIBZHL_API IsValidGridIndex(lua_Integer index, bool includeWalls = true) const;
```
Sylmir: I'm probably the one who used lua_Integer here and it should be just int, no need for a type dependency on Lua

repentogon/Patches/ModsMenuStuff.cpp
```diff
@@ -46,7 +46,7 @@ unordered_map<int, int> lastKeyPressTimeMap;
 bool cursorblink = false;
 float prevscroll = 0;
 int prevselect = 0;
-int minmods = 15;
+const unsigned int minmods = 15;
```
Sylmir: Why const exactly ? This feels like a workaround for a warning rather than a semantic choice

namishere: const is an intentional semantic choice as it never changes.

namishere: reverted const, organized variable definition block a bit

libzhl/functions/RoomConfigRoom.zhl
namishere: uses of lua_Integer for the static functions have been converted to int

libzhl/functions/Music.zhl
namishere: Added _minBound and _maxBound to Music, moved function over to IsaacRepentance_static.cpp

namishere: Realized now this isn't quite what you meant. Going to try documenting this class a bit

namishere: Now has a proper class, with a vector for it added to Music and used for this function

PR #428, opened 2024-04-26
Some function signatures and offsets I found

Guantol-Lemat:
- Rewind States
- Sold Items
- Curated Daily Challenges
- General Level Generator Data


wofsauge:
i cant get these changes to build... something seem to not be correct in the PR.

It probably has structure elements defined that are too big ( like your "SoldItems" array, which is too big to fit).

thru my testing, i couldnt find out how to fix the crashes, so maybe someone else can find out whats wrong here.

jsgnextortex:
Can confirm, didnt build for me either but at a glance I wasnt able to tell why either.

PR #438, opened 2024-05-03
No pointers?

Goganidze:
It's annoying that you can't put a sprite directly into the beam, and you have to store it. This should fix that


namishere:
Hi, very sorry that it's taken me this long to review this. I've decided that it would be for the best to go back to storing a dedicated Sprite object, like how it was handled pre-release. That should accomplish the same and better effect as this, with the tradeoff that changes made to the Sprite provided by the constructor won't be reflected in the Beam itself. Thank you very much!

PR #486, opened 2024-06-15
MC_POST_CYCLE_RATE_CALCULATE

Guantol-Lemat:
Also added previous additions to the Changelog


changelog.txt
ConnorForan, line 1: Remove unrelated changelog additions from this PR (if they are still missing from the changelog at head, feel free to just push them directly, though idk with how old this PR is). I guess even the changelog for MC_POST_CYCLE_RATE_CALCULATE will need to be moved. Or omit the changelog from the PR entirely and push something in afterwards.

repentogon/resources/scripts/enums_ex.lua
```diff
@@ -173,6 +173,7 @@ ModCallbacks.MC_PLAYER_INIT_PRE_LEVEL_INIT_STATS = 1127
 ModCallbacks.MC_PRE_NEW_ROOM = 1200
 ModCallbacks.MC_PRE_MEGA_SATAN_ENDING = 1201
 ModCallbacks.MC_POST_MODS_LOADED = 1210
+ModCallbacks.MC_POST_CYCLE_RATE_CALCULATE = 1211
```
ConnorForan, line 176: No action needed, just noting that this ID is still available

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1084,3 +1084,57 @@ void ASMPatchPrePlayerGiveBirth() {
...
+
+	cycleRate = (uint32_t)lua_tointeger(L, -1);
+	if (cycleRate == 0) { // Since the game uses this number as a divisor a value of 0 would cause a Crash
+		cycleRate = 1;
+	}
+	return cycleRate;
+}
+
+void ASMPatchPostCycleRateCalculate() {
+	SigScan scanner("837f??0075??c787????????00000000");
```
ConnorForan, line 1122: Update to new signature standard (ASM.zhl)

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1084,3 +1084,57 @@ void ASMPatchPrePlayerGiveBirth() {
...
+	}
+
+	lua_State* L = g_LuaEngine->_state;
+	lua::LuaStackProtector protector(L);
+	lua::LuaCaller caller(L);
+
+	lua_rawgeti(L, LUA_REGISTRYINDEX, g_LuaEngine->runCallbackRegistry->key);
+
+	lua::LuaResults failure = lua::LuaCaller(L).push(callbackid)
+	.push(collectible->_cycleCollectibleCount)
```
ConnorForan, line 1104: nit: fix indentation

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1084,3 +1084,57 @@ void ASMPatchPrePlayerGiveBirth() {
 	printf("[REPENTOGON] Patching Entity_Player::TriggerHeartPickedUp at %p for MC_PRE_PLAYER_GIVE_BIRTH_IMMACULATE\n", immaculateAddr);
 	PreBirthPatch(immaculateAddr, false);  // Immaculate
 }
+
+// MC_POST_CYCLE_RATE_CALCULATE(1211)
```
ConnorForan, line 1088: Just for ref, does this run one every update, or only each time the item "cycles"? Does the timer need to land exactly on the cycle rate, or does the internal code do a >(=) check (mostly just considering the possibility of changing the rate causing it to get stuck permanently. I have not looked at the vanilla code)

Also, does this ONLY run when the pickup is actually a cycling collectible?

Guantol-Lemat:
Since:
1. You can already get a similar effect with the MC_PRE_PICKUP_MORPH callback.
2. The cycle logic has been updated in REP+.
3. This was my first ASM patch.

I think it makes more sense to close this and revisit the idea later, if it’s still needed.

PR #488, opened 2024-06-19
Fixed problems caused by the compatibility wrapper

Guantol-Lemat:
- Seed and Decrease are no longer swapped.
- Flags is now passed to the function.


repentogon/LuaInterfaces/LuaItemPool.cpp
```diff
@@ -31,9 +31,8 @@ LUA_FUNCTION(Lua_ItemPoolGetCardEx)
 LUA_FUNCTION(Lua_ItemPoolGetCollectibleEx) {
 	ItemPool* itemPool = lua::GetUserdata<ItemPool*>(L, 1, lua::Metatables::ITEM_POOL, "ItemPool");
 	int poolType = (int)luaL_checkinteger(L, 2);
-	uint32_t seed = (unsigned int)luaL_optinteger(L, 3, Isaac::genrand_int32());
-	if (seed == 0) seed = 1;  // IDC
-	bool keepWeight = luaL_optinteger(L, 4, 1); //Fun Fact: main.lua creates a wrapper for GetCollectible that inverts the boolean value passed as Decrease, then passes it as an integer.
+	bool decrease = lua::luaL_optboolean(L, 3, false);
+	uint32_t seed = (unsigned int)luaL_optinteger(L, 4, Isaac::genrand_int32());
```
ConnorForan, line 35: Can `Isaac::genrand_int32()` return 0? I know that weirdly enough, the original wrapper for GetCollectible would use `Random()` as the default value for the seed, which I'm pretty sure can return zero and my understanding was that it was the same thing as `Isaac::genrand_int32()`.

Considering passing 0 as the seed always crashes the game perhaps we could just avoid that. I know my method of catching it was really lazy, but still.

There also doesn't feel like much need in allowing mods to pass 0 for the seed from the lua side knowing it will just crash the game. Maybe we could make that an error.

repentogon/LuaInterfaces/LuaItemPool.cpp
```diff
@@ -31,9 +43,11 @@ LUA_FUNCTION(Lua_ItemPoolGetCardEx)
...
+	if (seed == 0) {
+		seed = GetValidRNGSeed();
```
ConnorForan: Please just use 1, the effect to the RNG here is negligible. It's not worth the extra complexity of anything involving a loop.

Guantol-Lemat: Updated

ConnorForan:
Thanks!

Issue #222, opened 2024-01-05
Move function for ASMPatch


namishere:
I am writing a generic patch to be used multiple times with various different registers that requires me to `mov [destReg], [srcReg]`. The current `Move[x]` functions that have a register as the destination are in the form of `mov dword ptr ds:[destReg], srcReg`, which does not work for this patch and crashes. I would like a `ASMPatch.Move(ASMPatch::Registers srcReg, ASMPatch::Registers destReg)` that equates to `mov [destReg], [srcReg]`. Thank you!

Sylmir:
I didn't realize this what was you wanted to do at first. On x86 memory-to-memory moves are not possible. If you want to do `*eax = *ebx`, you need to do `ecx = *ebx; *eax = ecx;`.

Sylmir:
Update: register-to-register copy is requested, not memory-to-memory

namishere:
`CopyToRegister` does what this issue is requesting. Marking resolved.

PR #506, opened 2024-07-18
MC_GET_STATUS_EFFECT_TARGET

namishere:
Called when a status effect is about to be applied to an Entity. Allows returning another Entity to apply the status effect to instead. Will still attempt to recursively copy the status effect to all children, so best used with desired Parent.

Usage:
```lua
mod:AddCallback(ModCallbacks.MC_GET_STATUS_EFFECT_TARGET, function (_, entity)
	print(entity.Type)
	return Isaac.GetPlayer()
end, EntityType.ENTITY_MINISTRO)
```


guwahavel:
Could someone review this at some point so it could be merged? Not a huge urgency but this would make coding segmented enemies way easier and less buggy/jank.

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
```diff
@@ -0,0 +1,169 @@
...
+Entity* __stdcall GetStatusEffectTargetTrampoline(Entity* target) {
+	return target->GetStatusEffectTarget();
+}
+
+/* /////////////////////
+// Generic inline patch
+*/ /////////////////////
+
+void ASMPatchInlinedGetStatusEffectTarget(void* addr, ASMPatch::Registers entityReg, ASMPatch::Registers targetReg, ASMPatch::SavedRegisters::Registers savedReg, unsigned int jumpOffset) {
+	ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::GP_REGISTERS + ASMPatch::SavedRegisters::XMM1 - savedReg, true);
```
Sylmir: Use | instead of +, use & and ~ instead of -

A & ~B -> Keep all bits of A that are not set in B

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
```diff
@@ -0,0 +1,169 @@
...
+	ASMPatch patch;
+	patch.PreserveRegisters(savedRegisters)
+		.Push(entityReg)
+		.AddInternalCall(GetStatusEffectTargetTrampoline)
+		// poor man's mov reg,reg
```
Sylmir: What does that comment mean ? Why would you do mov reg, reg ?

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/* /////////////////////
+// Run patches
+*/ /////////////////////
+
+void PatchInlinedGetStatusEffectTarget()
+{
+	ZHL::Logger logger;
+	for (StatusEffectPatchInfo i : patches) {
```
Sylmir: Add const& to capture by reference, will improve perf and avoid unexpected stuff like weird copies

libzhl/functions/Entity.zhl
```diff
@@ -11,29 +11,12 @@ static void Entity::PlaySound(int ID, float Volume, int FrameDelay, bool Loop, f
...
-
-"558bec83ec30a1????????33c58945??56578b7d??8bf157e8????????84c00f85????????576afd":
-__thiscall void Entity::AddWeakness(const EntityRef& ref, int duration);
 
 "558bec5156ff75":
 __thiscall void Entity::AddIce(const EntityRef& ref, int duration);
 
 "558bec83e4f85156ff75??8bf1e8":
-__thiscall void Entity::AddKnockback(const EntityRef& ref, const Vector& pushDirection, int duration, bool takeImpactDamage);
+__thiscall void Entity::AddKnockback(EntityRef& ref, const Vector& pushDirection, int duration, bool takeImpactDamage);
```
Sylmir: Any reason why AddKnockback does not take a const reference to the EntityRef ?

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
```diff
@@ -0,0 +1,169 @@
...
+		.RestoreRegisters(savedRegisters)
+		.AddRelativeJump((char*)addr + jumpOffset);
+	sASMPatcher.PatchAt(addr, &patch);
+}
+
+/* /////////////////////
+// Thank you, God bless you, and God bless the United States of America.
+*/ /////////////////////
+
+bool avoidIceCrash = false;
```
Sylmir: Please add a comment in the code explaining why we need this hack, because right now I'm terrified of what releasing this feature will do

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
```diff
@@ -0,0 +1,169 @@
...
+		if (!IgnoreEffectFromFriendly(&ref)) {
+			if (GetStatusEffectTarget()) {
+				Entity* parent = this->_parent;
+				Entity* otherEnt = parent;
+				while (otherEnt = parent, otherEnt != nullptr) {
+					entity = otherEnt;
+					parent = otherEnt->_parent;
+				}
+			}
+			if (duration != 0 && ((entity->_flags & 1) == 0 || (entity->_flags & 0x2000000000000) != 0)) {
```
Sylmir: Please add 0x2000000000000 as a define somewhere, I have no idea what it represents. Sames goes for the 1 that you use in the first part of the if

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
ConnorForan: Perhaps they were not aware of `.CopyRegister(targetReg, ASMPatch::Registers::EAX)` and needed a way to move the value from EAX to whatever `targetReg` is :P

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
```diff
@@ -0,0 +1,169 @@
...
+				entity->_knockbackDirection = pushDirection;
+				if (10.0f < entity->_mass && ((this->_flags & 0x2000000000000) == 0)) {
+					float logMass = 1.0f / std::log10(entity->_mass);
+					entity->_knockbackDirection *= logMass;
+					int countdown = (int)(entity->_knockbackCountdown * logMass);
+					if (countdown < 1)
+						countdown = 1;
+					entity->_knockbackCountdown = countdown;
+				}
+				entity->_flags |= 0x800000000000;
```
Sylmir: Same as above for the 0x800000000000 flag

Sylmir (changes requested):
Please add defines instead of using hardcoded values, and add comments about this "ice crash" stuff. Right now you're the only person able to debug this if it breaks because we do not have the relevant information available

namishere:
Will address these later today

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
namishere: What we really need is to port over our Ghidra enums. It doesn't seem like there's an immediate way to export them, though...

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
namishere: I was not aware of `CopyRegister`. If this works how I think it does, this would close #222.

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
namishere: One specific call to AddKnockback when colliding with iced enemies was somehow leading to ECX containing 0x0 when calling GetType, which of course doesn't work well. I do agree in retrospect that this solution is a bit maniacal. I will try to work around it a different way.

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
namishere: I based this off of prior usage. I can correct this, but other patches will need to be re-evaluated.

libzhl/functions/Entity.zhl
namishere: Probably cheap solution to having trouble with my AddKnockback hook failing signature checks on game launch. I should be able to revert this.

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
namishere: Resolved by virtue of no longer reimplementing AddKnockback.

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
namishere: A more proper fix has been found and implemented.

repentogon/Patches/ASMPatches/ASMStatusEffects.cpp
namishere: This usage of it has been fixed. Further evaluation of patches pending.

PR #517, opened 2024-08-07
Repentogon socket

elliottcarlson:


jsgnextortex:
Allowing socket without luadebug is far too risky for the modding community and while allowing luadebug, all the api restrictions go out of the window anyway, by that point you can run any lua you want and do anything you want, even without repentogon....for this reason, luadebug, and its functionalities, were never really a focus for the project.
In other words, while the whole point of repentogon is to extend the api without needing to use luadebug, the socket library is out of scope and far too much of a security nightmare to consider....so this PR is unlikely to be merged, if I need to be honest. BUT, take in mind that REPENTOGON allows dll mods and you could tackle it that way, as an example theres catinsurance's dicord rich pressence mod, if you want to take a look:
https://github.com/catinsurance/Repentogon-Discord-Rich-Presence

elliottcarlson:
Had an itchy trigger finger on this PR -- #518 has a PR description. Will close this one with comment.

@jsgnextortex This does not remove the need for --luadebug - it just fixes it when --luadebug is enabled.

PR #529, opened 2024-08-22
Cards EX

Guantol-Lemat:
- Added Hidden field to Cards
- Added Weight filed to Cards
- Added Custom Availability Condition to Cards.


ConnorForan:
So, in regards to the custom availability conditions, since that involves holding onto a reference to a lua function on the C++ side

What happens when the mod that added that condition reloads, via luamod? Does the function get garbage collected, or does it get maintained by the C++ side?

Sylmir:
When you create a Lua-C reference to an object O through luaL_ref, you get an integer R. R is a key in the Lua registry, and O is associated to R in the registry, increasing its reference count by 1. This means the object cannot be garbage collected until the reference is freed (with luaL_unref). 

In the context where you get a reference to a function, I don't know how Lua preserves the environment of the function, in particular if it acts as a closure.

ConnorForan:
Yeah, that was one of my concerns, if this function reference existing could potentially preserve a bunch of a mod's tables in memory between luamods (due to preserving the environment of the function), as that could both potentially lead to the function no longer working as expected, or just simply taking up lua memory by preserving a mod's table/data.

True, you could expect mods to also re-set the condition function on reload anyway, which would circumvent the issue (assuming we do indeed free the reference from our side properly), but this still feels like a valid thing to consider with this.

Sylmir:
I'm writing a small test to see what happens, I'll keep you updated

Sylmir:
After some checks, it is possible to leak memory with this code, so I suggest putting a small disclaimer in the documentation. You need to be looking for the issue though, by registering a function that is actually a closure inside another function. So I'd say it's a non issue

ConnorForan:
Yeah, after chatting with Sylmir a little bit, this is probably not a big concern, there is the possibility of leaks or weird behaviour if these types of function references are mismanaged but (at least for this situation) it could likely only be a problem for the mod responsible, and probably would not cause any real issues otherwise. I'm not too worried about it anymore, just wanted to make sure we understood what could happen.

A couple of things I will note though:

- It seems like right now if the lua code errors, it will make the card not available due to returning false? Maybe it should return true instead so that we act as if the custom condition function was not there at all (and fall back to just the basegame's IsAvailable)

- The current changes are adding the changelogs to 1.0.11c, which was already released. I did a push today that created a new section in the changelog for whatever the next update will be, so you should sync up to head again and move your changelogs up to 1.x.x

PR #538, opened 2024-09-03
Reworked MC_POST_PICKUP_SELECTION

Guantol-Lemat:
- Passes requiredVariant, requiredSubType and rng as additional parameters
- Allows an additional 3rd argument to be returned in order to specify wether or not the callback should proceed as an Additive Callback or return the result early like in vanilla.


repentogon/resources/scripts/main_ex.lua
```diff
@@ -447,7 +447,7 @@ local typecheckWarnFunctions = {
 		["table"] = checkTableSizeFunction(4)
 	},
 	[ModCallbacks.MC_POST_PICKUP_SELECTION] = {
-		["table"] = checkTableSizeFunction(2)
+		["table"] = checkTableTypeFunction({ "integer", "integer", "boolean"})
```
ConnorForan, line 450: Does this trigger a warning if you don't provide the boolean?

Guantol-Lemat: No, it doesn't, tough now that I've checked it seems that neither this, nor the previous function, actually trigger a warning if you pass less than the specified arguments, only if you pass more (and in the current case if you pass an argument that does not match the specified type).

PR #545, opened 2024-09-12
Fixed -1 runeChance in reworked GetCardEx

Guantol-Lemat:
- Properly handle the case of -1 runeChance in the new GetCardEx
- Removed a duplicate operation that added RUNE_SHARD to the RunePool


Sylmir:
Looks good to me

PR #551, opened 2024-09-26
PRE/POST_ENTITY_SET_COLOR

Syntax-Sculptor:


repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -4313,21 +4313,21 @@ HOOK_METHOD(ItemPool, ForceAddPillEffect, (int32_t ID)->int) {
...
-	};
+	lua::LuaResults result = lua::LuaCaller(L).push(callbackid)
+		.pushnil()
+		.push(ID)
+		.push(ret)
+		.call(1);
+};
 
-	return ret;
+return ret;
```
ConnorForan: What's with the diffs up here? Accidental?

repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -4392,6 +4392,58 @@ HOOK_METHOD(Game, RestoreState, (GameState* gameState, bool startGame) -> void)
...
+			.push(fadeOut)
+			.push(share)
+			.call(1);
+
+		if (!result) {
+			if (lua_isboolean(L, -1) && !lua_toboolean(L, -1)) {
+				return;
+			}
+			else if (lua_isuserdata(L, -1)) {
+				color = lua::GetUserdata<ColorMod*>(L, -1, lua::Metatables::COLOR, "Color");
```
ConnorForan: @Sylmir , I'm curious about some specifics of how memory gets managed when we receive userdata from lua like this.

If someone initializes a new Color object from lua DURING this callback and returns it, could that Color get garbage collected since there are no more references to it on the lua side? Maybe it persists so long as its still on the lua stack where we got it from, but if so how long is it safe to use? Within the scope of the current LuaStackProtector or something?

My concern is whether it is unsafe to receive a new Color pointer from lua here, and then pass it along to super and another callback below.

Maybe it is fine but I wanna understand who's maintaining ownership of this object in this case.

repentogon/LuaInterfaces/CustomCallbacks.cpp
Syntax-Sculptor: > What's with the diffs up here? Accidental?

oh. i think visual studio's formatter messed  up. i will fix it.

repentogon/LuaInterfaces/CustomCallbacks.cpp
Sylmir: Objects on the stack still have a reference pointing towards them, the stack itself. I don't have a definitive proof of that, but I can make the following observation : luaL_newtable (and lua_newtable) both create a new table that is pushed on the stack, and tables are subject to garbage collection. I'm also expecting Lua to handle this case in a more natural way than, say, C or (especially) C++ would. 

As to the validity of the object, we need to distinguish two things that can be considered the object here : the userdata itself, and the object represented by the userdata. What I mean here is the following: what is the type of the userdata ? Is the userdata a Color*, as-in we allocated a full Color object, therefore the size of the userdata is sizeof(Color) ? Or is it a Color**, as-in we allocated a Color through new / malloc and stored the returned pointer inside a userdata ? 

If the userdata is a full object, a Color, then the validity of the Color itself is bound by the presence of the userdata on the stack. As soon as the userdata is no longer on the stack, you can no longer safely use the Color.

If the userdata is a reference to a Color, i.e. the userdata stores a pointer to a Color, then the Color itself is valid until you free it. The userdata is no longer safe to use because its content may change if memory gets reused afterwards, but the pointer and the Color pointed by the pointer remain valid.

Sylmir: Considering the LuaResults is declared inside this if, the stack will get cleaned up when we exit this if, so be careful. In particular the value of `color` in the next if will be in an unstable state considering it may or may not have been changed by this part of the function

ConnorForan: Thanks for the insight, Sylmir!

@Syntax-Sculptor I think we should make a local copy of the color we get back from lua, in case the color we got back is unsafe to use after the lua stack is cleaned up.

If an override color was provided, we pass a pointer to that to super and to the post callback. Otherwise we continue to use the original color.

PR #559, opened 2024-10-06
vanilla tweaks: add a launch flag to skip archived file checksumming

im-tem:
This pull request provides a new `-skipchecksum` launch parameter which makes the game skip a block of code for performing checksum generation and validation, leading to a massive startup time boost. It is opt-in to avoid potential issues that could arise from a checksum mismatch.
Archive read speed comparison on my machine with NVMe SSD (using times provided by the log entry, avg over three consecutive launches):
without the flag - (8930+9035+7869)/3 -> 8611ms
with the flag - (1270+1265+1262)/3 -> 1266ms

Shaving seven whole seconds is rather huge I'd say.


im-tem:
did anyone test this? should be safe to merge but still

PR #561, opened 2024-10-07
ImGuiWindowFlags

BabyblueSheep:
Adds `ImGui.GetWindowFlags(string WindowId)` and `ImGui.SetWindowFlags(string WindowId, int flags)` for adding window flags to windows.
Additionally, adds an `ImGuiWindowFlags` enum and a window + number input in `test_imgui.lua` for testing window flags.


wofsauge (approved):
looks good to me

PR #567, opened 2024-10-14
Fix crashes involving status effect functions

namishere:
This does a few things:
1. The prototypes of `AddConfusion` and `AddSlowing` have been fixed. `AddConfusion` was missing a bool, leading to a stack corruption. `AddSlowing` had a uint that should have been a float, and the `ColorMod*` is now a proper (?) full `ColorMod` object.
2. I went back through the functions and noticed they all have the `Entity*` in `ESI` right before the `GetStatusEffectTarget` un-inlining patch. All patches have been changed to use this register for `entityReg`. A couple of signatures have additionally been slightly changed to be consistent with where the other patches were hooking. This is mostly for consistency, but it's an improvement over before.
3. Some extra clarification has been added as to how these patches work, for the sake of future maintainers.

These crashes don't manifest themselves in a standalone release of REPENTOGON, but other ZHL mods attempting to hook or call them would.


ConnorForan (approved):
Looks good to me, thanks for the fixes + renaming

PR #577, opened 2024-10-26
imgui: downgrade to opengl2 impl

im-tem:
This branch contains a commit that replaces the OpenGL 3 implementation of Dear ImGui with its' OpenGL 2 counterpart. May fix rendering issues on old GPUs. Not tested beyond confirmation of it functioning on OpenGL 4.6 capable hardware.


im-tem:
Creating a new rendering context may not be the best approach, it would be better to find the aspect of the game's context that messes with imgui's rendering and try resolving that instead.

im-tem:
this commit tries an entirely different approach under the assumption that a system stuck with opengl 2.x will not be able to access any opengl 3.x+ attributes, implying that opengl 2 imgui should render fine. may be wrong but worth trying on real hardware.

im-tem:
Superseded by https://github.com/TeamREPENTOGON/REPENTOGON/commit/a9fd33458251b039c80f65693a9b5e6e1fe54a68

PR #582, opened 2024-11-06
Updated Documentation with all missing functions currently available on the main branch

wofsauge:
Someone should maybe look over the changes and see if anything should be removed or changed


docs/docs/ItemConfig.md
Guantol-Lemat, line 1: Shouldn't most of these be moved to the appropriate ItemConfig file?

ConnorForan: Most of these ARE already documented on their appropriate class (ItemConfig_Item or ItemConfig_Card)

They should not be added here.

ConnorForan:
There seems to be multiple issues with this PR, was this auto generated?

To name a few...

There is already a page for GenericPrompt: https://repentogon.com/GenericPrompt.html
Already-documented functions for ItemConfig_Card and ItemConfig_Pill were incorrectly added to ItemConfig
Already-documented functions for EntitiesSaveState were incorrected added to EntitiesSaveStateVector

There are probably other similar cases, there seems to be an issue with how these were generated that needs to be addressed

jsgnextortex:
No, this PR was made to check Wof's output on the functions that were left undocumented in the code. If you mean the branch, the point of the branch is to hold doc changes that are tied to features that are not yet released to the public and they should be merged when these features reach the public (as in, if theres a major version), The reason why its been a while since the last merge/rebase is because its been a while since the last major version.

jsgnextortex:
Now, a lot of the proposed changes on this PR respond to allegedly undocumented stuff thats supposedly already available to the public, so those cases could be pushed straight to main on these cases.

docs/docs/BossPool.md
```diff
@@ -24,6 +24,18 @@ ___
 ### GetName () {: aria-label='Functions' }
 #### string GetName ( ) {: .copyable aria-label='Functions' }
 
+___
+### GetPool () {: aria-label='Functions' }
+#### [BossPool](BossPool.md) GetPool ( ) {: .copyable aria-label='Functions' }
```
jsgnextortex: These are actually all exposed under BossPoolManager

docs/docs/EntitiesSaveStateVector.md
```diff
@@ -13,10 +13,166 @@ ___
...
+### SetType () {: aria-label='Functions' }
+#### void SetType ( int type ) {: .copyable aria-label='Functions' }
+
+___
+### SetU1 () {: aria-label='Functions' }
+#### void SetU1 ( int num ) {: .copyable aria-label='Functions' }
+
+___
+### SetVariant () {: aria-label='Functions' }
+#### void SetVariant ( int variant ) {: .copyable aria-label='Functions' }
```
jsgnextortex: Theses are normally documented on EntitiesSaveState, and they seem to be all documented already

jsgnextortex:
Peeled up most of what I could find that was out of place, if someone can do another pass at whats left, it would be appreciated....but I think that should be it for the most part.

docs/docs/EntityPlayer.md
```diff
@@ -190,6 +190,10 @@ ___
 ### ClearQueueItem () {: aria-label='Functions' }
 #### void ClearQueueItem ( ) {: .copyable aria-label='Functions' }
 
+___
+### DischargeActiveItem () {: aria-label='Functions' }
```
ConnorForan: This vanilla function was re-hooked to add a safety check and did not actually have its functionality changed.

Therefore I don't think it should be documented here.

This also applies to FullCharge.

ConnorForan (commented):
This PR does contain a few things that could have been changed at head instead of on this branch. Including but potentially not limited to:

- EntityNPC:GetFlyingOverride exists in release and was just undocumented
- The various changes from "bool" to "boolean"
- Formatting fixes in ItemConfig.md
- Fix and default args for Room:GetItemPool

Now, personally I don't care enough to press if you don't feel like removing those from this PR and pushing those changes to main. Just wanted to point them out.

The rest of the changes lgtm

PR #624, opened 2025-01-30
[Rep+] Reload Custom Shaders

Lunastela:
As the title implies, adds the ability to reload custom shaders via the use of the reloadshaders command.

https://github.com/user-attachments/assets/e17af34c-f66b-4aa7-a05d-ce00a7d72fe4

(Please ignore the mod.)

### DIFF
--- a/repentogon/Patches/Anm2Extras.cpp
+++ b/repentogon/Patches/Anm2Extras.cpp
@@ -142,10 +142,13 @@ std::string FindShaderFullPath(const std::string& path) {
 }
 
 // Abstracted shader loading (for reloading shaders via console)
-void LoadCustomShader(const std::string& path, KAGE_Graphics_Shader* shader, bool champion) {
+bool LoadCustomShader(const std::string& path, KAGE_Graphics_Shader* shader, bool champion) {
 	const std::string fullpath = FindShaderFullPath(path);
+	if (fullpath.empty())
+		return false;
 	KAGE_Graphics_VertexAttributeDescriptor* desc = champion ? &g_ColorOffset_Champion_VertexAttributes : &g_ColorOffset_VertexAttributes;
 	KAGE_Graphics_Manager_GL::LoadShader(shader, desc, fullpath.c_str());
+	return true;
 }
 
 // Returns the custom shader corresponding to the relative path.
@@ -156,9 +159,11 @@ CustomShader* GetOrLoadCustomShader(const std::string& input_path, const bool ch
 	const std::string path = NormalizeShaderPath(input_path);
 	auto& shader_map = champion ? custom_champion_shaders : custom_shaders;
 	if (shader_map.count(path) == 0) {
-		if (FindShaderFullPath(path).empty())
+		CustomShader customShader;
+		customShader.path = path;
+		if (!LoadCustomShader(path, &customShader.shader, champion))
 			return nullptr;
-		LoadCustomShader(path, &shader_map[path].shader, champion);
+		shader_map[path] = std::move(customShader);
 	}
 	return &shader_map[path];
 }
--- a/repentogon/Patches/Anm2Extras.cpp
+++ b/repentogon/Patches/Anm2Extras.cpp
@@ -141,6 +141,7 @@ std::string FindShaderFullPath(const std::string& path) {
 	return "";
 }
 
+// Abstracted shader loading (for reloading shaders via console)
 void LoadCustomShader(const std::string& path, KAGE_Graphics_Shader* shader, bool champion) {
 	const std::string fullpath = FindShaderFullPath(path);
 	KAGE_Graphics_VertexAttributeDescriptor* desc = champion ? &g_ColorOffset_Champion_VertexAttributes : &g_ColorOffset_VertexAttributes;
@@ -150,12 +151,13 @@ void LoadCustomShader(const std::string& path, KAGE_Graphics_Shader* shader, boo
 // Returns the custom shader corresponding to the relative path.
 // Initializes the shader if necessary.
 CustomShader* GetOrLoadCustomShader(const std::string& input_path, const bool champion) {
-	if (input_path.empty()) {
+	if (input_path.empty())
 		return nullptr;
-	}
 	const std::string path = NormalizeShaderPath(input_path);
 	auto& shader_map = champion ? custom_champion_shaders : custom_shaders;
 	if (shader_map.count(path) == 0) {
+		if (FindShaderFullPath(path).empty())
+			return nullptr;
 		LoadCustomShader(path, &shader_map[path].shader, champion);
 	}
 	return &shader_map[path];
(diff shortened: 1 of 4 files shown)


repentogon/Patches/Anm2Extras.cpp
```diff
@@ -145,24 +141,24 @@ std::string FindShaderFullPath(const std::string& path) {
 	return "";
 }
 
+// Abstracted shader loading (for reloading shaders via console)
+void LoadCustomShader(const std::string& path, KAGE_Graphics_Shader* shader, bool champion) {
+	const std::string fullpath = FindShaderFullPath(path);
```
Sylmir, line 146: Please add a sanity check to avoid loading a shader with an empty path. The function LoadCustomShader may return a boolean indicating whether it failed or not, or an error code so that you may log something in the console or write to a log file (depending on call context)

repentogon/Patches/Anm2Extras.cpp
```diff
@@ -145,24 +141,24 @@ std::string FindShaderFullPath(const std::string& path) {
...
-		if (fullpath.empty()) {
+		if (FindShaderFullPath(path).empty())
 			return nullptr;
-		}
-
-		// Initialize the CustomShaderInfo object and contained KAGE_Graphics_Shader.
-		shader_map[path].path = path;
-		KAGE_Graphics_VertexAttributeDescriptor* desc = champion ? &g_ColorOffset_Champion_VertexAttributes : &g_ColorOffset_VertexAttributes;
-		KAGE_Graphics_Manager_GL::LoadShader(&shader_map[path].shader, desc, fullpath.c_str());
+		LoadCustomShader(path, &shader_map[path].shader, champion);
```
Sylmir: You compute FindShaderFullPath twice, once here and a second time in LoadCustomShader. Overload LoadCustomShader to have a version that computes the full path and one that takes the full path as parameter

repentogon/Patches/Anm2Extras.h
```diff
@@ -13,4 +13,14 @@ bool HasCustomShader(ANM2* anm2, const std::string& path, const bool champion);
...
 
-void ASMPatchesForANM2Extras();
+struct CustomShader {
+	std::string path;  // Relative path starting from `.../resources/`, without the file extensions, ie `shaders/myshader`
+	KAGE_Graphics_Shader shader;
+};
+
+void LoadCustomShader(const std::string& path, KAGE_Graphics_Shader* shader, bool champion);
+
+extern std::unordered_map<std::string, CustomShader> custom_shaders;
```
Sylmir, line 23: I may be missing something, but when are you writing something in either of these maps ? I see them used in read operations (mainly count() and ranged for loops), but I never see insert, [] or anything used to actually insert something in them

repentogon/Patches/Anm2Extras.cpp
ConnorForan: I think it'd be easier to just not call FindShaderFullPath here at all and, as per your other comment, just have LoadCustomShader return a boolean or something if the path is invalid. Inside LoadCustomShader is the only place you need to compute the full path.

repentogon/Patches/Anm2Extras.h
ConnorForan: The struct is automatically initialized with default values when accessed in std maps.

I'm fairly confident the original code at least would take care not to initialize entries by accident (IE, the check for `shader_map.count(path) == 0`)

Sylmir: Ah, right, completely missed the fact that KAGE_Graphics_Manager_GL::LoadShader takes a result pointer, not an input pointer. I'm going to say it's very sneaky and probably mandates to separate the initialization steps:

```cpp
CustomShader customShader;
customShader.path = path;
LoadCustomShader(path, &customShader.shader, champion);
shader_map[path] = std::move(customShader);
```

Not sure what the point of the path field of CustomShader is though

Also, side-note, we do not register constructors for KAGE objects most of the time (because we haven't found them), so relying on default initialization is very dangerous here. We should probably note somewhere to find a default constructor for KAGE_Graphics_Shader

repentogon/Patches/Anm2Extras.cpp
lizzyapp: Oops, didn't consider that when reloading a shader it may not exist. Working on it, thanks.

repentogon/Patches/Anm2Extras.h
lizzyapp: The check still exists, it's in the GetOrLoad function, the only other time this function is run is when the key still exists within the map. Would you like me to restructure the function to check the map itself instead? Didn't figure it'd be used for much else

lizzyapp: I don't believe path is used for anything. It was unset before any of these changes.

ConnorForan: If completely unused, the path field being part of the struct may be an artifact of some prior version of the code that was left in. It's possible the CustomShader struct could be replaced by just the KAGE_Graphics_Shader directly, but not strictly necessary for the PR.

Anyway, the main takeaway of this thread is to just make sure that elements in the map are not default-initialized unintentionally via [] access

Lunastela:
This seems to be working still, now with the suggested changes. Sorry, I am very inexperienced by the looks of it.

repentogon/Patches/ConsoleHooks.cpp
```diff
@@ -145,6 +146,18 @@ HOOK_METHOD(Console, RunCommand, (std::string& in, std::string* out, Entity_Play
...
     };
 
+    if ((in == "reloadshaders") || (in.rfind("reloadshaders ", 0) == 0)) {
+        if (!custom_shaders.empty()) {
+            for (auto & [ key, value ] : custom_shaders)
+                LoadCustomShader(key, &value.shader, false);
+        }
+        if (!custom_champion_shaders.empty()) {
+            for (auto & [ key, value ] : custom_champion_shaders)
+                LoadCustomShader(key, &value.shader, false);
```
ConnorForan: Need to pass "true" to the champion boolean for the champion shaders

repentogon/Patches/ConsoleHooks.cpp
```diff
@@ -145,6 +146,18 @@ HOOK_METHOD(Console, RunCommand, (std::string& in, std::string* out, Entity_Play
         return super(in,out,player);
     };
 
+    if ((in == "reloadshaders") || (in.rfind("reloadshaders ", 0) == 0)) {
+        if (!custom_shaders.empty()) {
```
ConnorForan: I guess the nonempty checks aren't strictly necessary, since the for each iterator will just not do anything if the map is empty.

Lunastela:
Unrelated to the PR (mostly), but I think it's okay to keep Path on CustomShaders. It would be neat to add something like GetCustomShader() which returns the path of a custom shader. It would be niche, but someone would get use out of it.

ConnorForan:
Mainly just redundant considering the map key is the exact same string as the path, heh. Not major either way but I might fiddle with it later.

In any case thanks for adding this feature! Hadn't even considered reloading these with the reloadshaders command honestly. Will merge this in later and give it a quick test myself later on if nothing else comes up.

PR #628, opened 2025-02-07
ItemPoolManager

Guantol-Lemat:
This adds:
- `ItemPoolManager`: A new class that manages vanilla and custom itempools' data.
- `AddCollectible`: A Lua method to add collectibles to an itempool permanently, as if they were defined within an .xml file.
- `AddTemporaryCollectible`/`RemoveTemporaryCollectible`: Two Lua methods to add/remove collectibles in an itempool only for the current run; these are automatically saved and restored by the manager. (only temporary collectibles can be removed)
- Several improvements to the previous implementation of CustomItemPools.


repentogon/Patches/ItemPoolManager.h
```diff
@@ -0,0 +1,220 @@
...
+	std::map<size_t, ItemPool*> m_PoolByName;
+
+public:
+	static ItemPoolManager& Get()
+	{
+		static ItemPoolManager s_Instance;
+		return s_Instance;
+	}
+
+	static void __Init() noexcept;
```
ConnorForan, line 59: What is the purpose of noexcept throughout?

Guantol-Lemat: Mainly to declare that the functions should never allow an exception leave their scope.
Also, it can enable the compiler to perform certain optimizations as a bonus.

ConnorForan:
Sounds good, thanks for the extra details. Will likely merge this in soon after another quick pass over

PR #647, opened 2025-03-29
Virtual Room Sets

Guantol-Lemat:
- Adds the `VirtualRoomSetManager` class to dynamically extend vanilla RoomSets at runtime without causing reallocations that would otherwise lead to crashes.
- Adds a Lua method `AddRooms` in both the **RoomConfig** and **RoomConfigSet** lua classes, that can convert a table of Lua Rooms into RoomConfig_Room objects, which are then added to the VIrtualRoomSetManger.


repentogon/Patches/VirtualRooms.cpp
```diff
@@ -0,0 +1,1376 @@
...
+
+template<typename RoomsContainer>
+inline std::vector<RoomConfig_Room*> VirtualRoomManager::__add_rooms_implementation(RoomsContainer&& rooms, LogUtility::LogContext& logContext, bool isRestoringRooms) noexcept
+{
+	static_assert(std::is_same_v<std::decay_t<RoomsContainer>, std::vector<RoomConfig_Room>> || std::is_same_v<std::decay_t<RoomsContainer>, std::vector<std::optional<RoomConfig_Room>>>, "Invalid RoomContainer for add_rooms implementation");
+	constexpr bool isOptional = std::is_same_v<std::decay_t<RoomsContainer>, std::vector<std::optional<RoomConfig_Room>>>;
+
+	auto insertionData = this->prepare_insertion_data(rooms, logContext, isRestoringRooms);
+
+#ifndef NDEBUG
```
ConnorForan: Can you explain the purpose of all of these NDEBUG blocks?

repentogon/Patches/VirtualRooms.cpp
```diff
@@ -0,0 +1,1376 @@
...
+	}
+}
+
+void VirtualRoomManager::RoomSet::insert_rooms(RoomSetInsertionData& roomSet, InsertionData& insertionData, bool isRestoringRooms) noexcept
+{
+	for (size_t i = 0; i < roomSet.roomData.size(); i++)
+	{
+		auto& roomData = roomSet.roomData[i];
+
+		assert(!roomData.mapEntry.IsInInvalidState());
```
ConnorForan: You have a lot of asserts in this file. These crash when tripped, right?

Are most of these really not anything we can continue or recover from?

repentogon/Patches/VirtualRooms.h
ConnorForan, line 1: Do you think you could make a pass over this header file and add some comments, such as some high-level notes and function comments? Doesn't have to be excessive, you've just got quite a bit here and it would help to make the implementation easier to ingest for others in the future.

repentogon/Patches/VirtualRooms.cpp
```diff
@@ -0,0 +1,1376 @@
...
+
+HOOK_METHOD(RoomConfig, GetRooms, (int stage, int type, int shape, int minVariant, int maxVariant, int minDifficulty, int maxDifficulty, unsigned int* doors, unsigned int subtype, int mode) -> RoomConfigRoomPtrVector)
+{
+	auto rooms = super(stage, type, shape, minVariant, maxVariant, minDifficulty, maxDifficulty, doors, subtype, mode);
+	VirtualRoomManager::__GetRooms(rooms, stage, type, shape, minVariant, maxVariant, minDifficulty, maxDifficulty, *doors, subtype, mode);
+
+	return rooms;
+}
+
+HOOK_STATIC(Level, write_room_config, (RoomConfig_Room* roomConfig, GameStateRoomConfig* gameStateRoom, std_deque_RoomConfigRoom* deque) -> void, __stdcall)
```
ConnorForan: Just confirming my understanding: These two Level hooks are where the game is normally saving/loading the info about which RoomConfigRoom is loaded for a floor's savedata, normally just being enough info for the game to fetch the room from the RoomConfig upon continue. Since virtual rooms are stored separately, the data you've stored for the virtual room is different than a regular room, and you can identify virtual rooms via a custom flag and take over fetching the RoomConfigRoom. This is kept consistent across save/load due to how you save/load the RoomConfigRooms to your json savedata.

ConnorForan (commented):
Wanted to try to confirm I understand a few of the high-level things here.

1. The VirtualRoomManager infrastructure virtually extends the RoomConfig lists by hooking the access functions to them. (Did you ever run into any cases where the game sidesteps these functions to access room data?)

2. The VirtualRoomManager itself owns the RoomConfigRooms created from lua, and is responsible for saving/loading json & GameState modifications to allow them to stay persistent across save/load, so they can be used in all the same ways as regular rooms.

3. Apart from the saving/loading, since we are creating legitimate RoomConfigRooms and the game primarily interacts with them via pointers, most vanilla interactions with the RoomConfigRoom just kinda work naturally.

4. If no virtual rooms are added, we aren't interfering with any of the vanilla game's behaviour/logic

repentogon/Patches/VirtualRooms.cpp
Guantol-Lemat: in C++ asserts are only enabled when the NDEBUG macro is not defined, so both the expression and the potential crash are omitted when compiling in Release mode.

I just made aggressive use of them to make sure that we are always in a valid state while testing; and they also are convenient as pseudo documentation for what constitutes a valid state/input.

repentogon/Patches/VirtualRooms.cpp
Guantol-Lemat: Since these functions are only used as a way to assert the state of the manager, and do not perform any additional state management, I used the NDEBUG macro to exclude the code from Release mode.

I'm certain the compiler could have noticed on it's own that all that code ultimately did not perform any meaningful operation, but I wanted to make sure.

repentogon/Patches/VirtualRooms.cpp
Guantol-Lemat: Regular virtual rooms are fully compatible with the way regular GameStateRoomConfig rooms are saved and restored. The reason for the hooks are "Restored" rooms (which are the virtual rooms built from the json saves) that are contained in their own "hidden" room set; and are, as such, incompatible with the vanilla GameStateRoomConfig encoding.
To notify that that room belongs to that specific set, I used a custom flag (which is completely unused by the game, based on Level::write_room_config, which is the function that encodes GameStateRoomConfig) so that when Level::read_room_config is used to decode the room I check if the room is a Restored room, and I fetch it from the hidden set.

Guantol-Lemat (commented):
1. Unless I missed some very obscure interaction, every function that interacts with RoomConfigRoom, including those from RoomConfig itself, end up passing through these two "main" functions.
2. Yes, the VirtualRoomManager essentially emulates all the management functions from RoomConfig, so **Adding Rooms**, **Guaranteeing Unique Room Ids** (which is done by interacting with the main RoomConfig's RoomSets), **Getting Rooms**, **Saving Rooms** and **Resetting Weights**.
The only missing interaction is Unloading (the only thing we do is reevaluating the room ids).
3. Yes, most of the meaningful interaction happen using the RoomConfigRoom pointers, and as such they work out of the box.
4. If no rooms are added then no room are added to the "result" of the Getter functions; so we can expect seed parity when no mod interacts with the manager.

repentogon/Patches/VirtualRooms.cpp
ConnorForan: Just curious, how come the restored rooms are treated differently than regular virtual rooms? I expected that the load from json would just add them to the VirtualRoomManager in the same way as when they were added originally? Or does that make it difficult to keep the rooms connected with the level/gamestate?

Guantol-Lemat: Restored rooms are designed to be a failsafe mechanism.
If a mod did not add the rooms that were previously being used in the level in time, then the room saved in the json file is added to the restored room set, and the GameStateRoomConfig is tied to this new entry, otherwise we link the room to the one from the regular set, and the generated room is discarded.
This is to keep the actual room sets only influenced by the current set of mods, while also preventing crashes from missing rooms, without "polluting" the room set with rooms that shouldn't exist yet.

ConnorForan: Okay, I get you know. Now that I think of it again, loading rooms from the json into the same sets would have easily led to duplicate rooms if mods load the same lua rooms again.

sgtm

ConnorForan (approved):
Thank you for answering my questions. The approaches taken to make this work look good to me.

ConnorForan:
Thank you for adding the extra comments and such.

Approved for my part, so I'm content with you merging this in whenever

PR #653, opened 2025-04-07
Update RoomConfigHolder.md

Blablabla3012:
`SubType` is wrong: to get the roomsubtype you need to write `theRoom.Data.Subtype`. no capital T

### DIFF
--- a/docs/docs/RoomConfigHolder.md
+++ b/docs/docs/RoomConfigHolder.md
@@ -14,7 +14,7 @@ tags:
 ## Functions
 
 ### GetRandomRoom () {: aria-label='Functions' }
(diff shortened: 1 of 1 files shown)


im-tem:
Functions in Lua do not have strong naming, the proposes names are merely suggestions as the argument's purpose is decided by its order
https://github.com/TeamREPENTOGON/REPENTOGON/blob/9460637bc710d6fc34ac35c9cfa9029a04ba296b/repentogon/LuaInterfaces/Room/LuaRoomConfig.cpp#L98

So, SubType being written instead of Subtype changes nothing (I assume this has happened because Entity has its' subtype field called SubType...)

Anyways, it'd probably a safe enough merge?

PR #688, opened 2025-07-07
Update Korean Localization

sawalk:
This is a first draft translation. There's still some parts that don't have unity, but it's all translated.


epfly6:
From korean mutual: '스타일로' 라는 말이 두 번 중복되어 나타나는 오류가 있으니 수정 바람
image

Issue #441, opened 2024-05-08
[Bug?] MC_PRE_CHANGE_ROOM return table likely not behaving as intended


drpandacat:
Returning an index other than the one of the current room you are in will cause a loop of constantly being sent to the same room over and over again, instead of being sent to the target room

```lua
mod:AddCallback(ModCallbacks.MC_PRE_CHANGE_ROOM, function (_, idx, dimension)
    local level = game:GetLevel()
    local new = level:GetRandomRoomIndex(false, Random() + 1)

    return {new, dimension}
end)
```

Sanio46:
This isn't what I described from my loose testing when you originally commented about this. Here's what I said:
- Returned TargetIdx appears to be completely ignored unless you're in the same room as your returned TargetIdx
- All but one door will return you to the same room. One door will have you exit, but...in an odd way that I can't accurately describe.

orrekorre:
ahh thanks for the clarification, i tried to explain what i found from the limited amount of testing i did

epfly6:
Should be fixed in https://github.com/TeamREPENTOGON/REPENTOGON/commit/b9fff5117bddfbed03e5124bde25ca913a32aa34

PR #699, opened 2025-07-29
Fix every issue related to PRE_LEVEL_CHANGE_ROOM

epfly6:


repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -1054,7 +1054,21 @@ HOOK_METHOD(Level, ChangeRoom, (int roomId, int dimension) -> void) {
 	if (!result) {
 		if (lua_istable(L, -1)) {
 			if (lua_rawlen(L, -1) == 2) {
-				super(lua::callbacks::ToInteger(L, 1), lua::callbacks::ToInteger(L, 2));
+				const int changedRoomId = lua::callbacks::ToInteger(L, 1);
+				const int changedDimension = lua::callbacks::ToInteger(L, 2);
+
+				RoomDescriptor* roomDesc = nullptr;
+
+				roomDesc = GetRoomByIdx(changedRoomId, changedDimension);
```
ConnorForan: Sorry this is very minor, but lines 1060~1062 can just be one line, right?

`RoomDescriptor* roomDesc = GetRoomByIdx(changedRoomId, changedDimension);`

repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -1054,7 +1054,21 @@ HOOK_METHOD(Level, ChangeRoom, (int roomId, int dimension) -> void) {
...
+				if (!roomDesc || !roomDesc->Data) {
+					logViewer.AddLog(LogViewer::Game, "MC_PRE_CHANGE_ROOM: Tried to switch to room with Idx %d and dimension %d, but it doesn't exist\n", changedRoomId, changedDimension);
+					super(roomId, dimension);
+				}
+				else {
+					g_Game->_leaveDoor = -1;
```
ConnorForan, line 1067: Just asking for context - what does this fix?

Guantol-Lemat: By setting leaveDoor to NO_DIRECTION, modders no longer have to set it manually—as any other value causes the intended gridIdx to be ignored and instead uses the gridIdx of leaveDoor. Since every warp in the original game sets leaveDoor to NO_DIRECTION, before changing the room, this behavior appears to be intentional.

epfly6: Supposedly should fix infinity warp if you're teleporting to the same room, mentioned in: https://github.com/TeamREPENTOGON/REPENTOGON/issues/441

repentogon/LuaInterfaces/CustomCallbacks.cpp
epfly6: True. Changed the line

PR #711, opened 2025-08-20
Native custom active item's GFX support

epfly6:
This pull request introduces a new "activegfx" property for the items.xml file. This attribute allows you to set a separate sprite for rendering the active item, similar to items such as the Jar, Urn of Souls, Smelter, and etc. To control the crop offset of a rendered sprite, you can now return "CropOffset" field in PRE_PLAYERHUD_RENDER_ACTIVE_ITEM callback (currently for modded items only).

All active slots should be supported

```xml
<!-- fetching sprite from gfxroot/collectibles/activegfx -->
<active id="1" name="Cain's Lost Key" description="Open your treasure" gfx="collectibles_sisterskey.png" maxcharges="4" activegfx="collectibles_theCovenant_custom.png"/>
```

### DIFF
--- a/repentogon/Patches/ASMPatches/ASMCallbacks.cpp
+++ b/repentogon/Patches/ASMPatches/ASMCallbacks.cpp
@@ -1598,9 +1598,9 @@ HOOK_METHOD(PlayerHUD, RenderActiveItem, (unsigned int activeSlot, const Vector&
 			cropOffset.x += !_player->_movingBoxContents.empty() ? 32.0f : .0f;
 			break;
 		case COLLECTIBLE_EVERYTHING_JAR:
-			
 			const int charges = std::max(_player->GetBatteryCharge(activeSlot), 12);
 			cropOffset.x += (float)(charges + 1) * 32.0f;
+			break;
 			
 		case COLLECTIBLE_FLIP:
 			cropOffset.x += _player->GetPlayerType() == ePlayerType::PLAYER_LAZARUS2_B ? 32.0f : .0f; //is dead Tainted Lazarus


repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1585,8 +1587,29 @@ HOOK_METHOD(PlayerHUD, RenderActiveItem, (unsigned int activeSlot, const Vector&
...
+	if (activeItemID < CollectibleType::NUM_COLLECTIBLES) {
+		switch ((CollectibleType)activeItemID) {
+		case COLLECTIBLE_MOVING_BOX:
+			cropOffset.x += !_player->_movingBoxContents.empty() ? 32.0f : .0f;
+			break;
+		case COLLECTIBLE_EVERYTHING_JAR:
+			/*
+			const int charges = std::max(_player->GetBatteryCharge(activeSlot), 12);
+			cropOffset.x += (float)(charges + 1) * 32.0f;
+			*/
```
ConnorForan: atm with no break, pretty sure this will fall through to the logic used for flip

Though, I do realize you mentioned elsewhere that only modded actives are supported atm. That is fine by me, though for now maybe comment out this entire block, add a comment here also noting that support for vanilla actives is currently disabled, AND if there is some particular issue with everything jar, please make a note of that as well for posterity (in case we end up leaving the support for vanilla actives disabled for some time)

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1625,6 +1650,7 @@ HOOK_METHOD(PlayerHUD, RenderActiveItem, (unsigned int activeSlot, const Vector&
 		}
 	}
 
+	activeRenderCropOffsetCache[_playerHudIndex][activeSlot] = cropOffset;
```
ConnorForan: Add a bound check for activeslot here just in case it is an invalid value

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1670,6 +1696,91 @@ void ASMPatchHideChargeBar() {
...
+	sASMPatcher.PatchAt(addr, &patch);
+}
+
+void __stdcall GetPlayerHUDActiveItemCropOffset(int activeSlot, PlayerHUD* playerHUD, float origCropX, float origCropY) {
+
+	float x = origCropX;
+	float y = origCropY;
+
+	//currently modify crop offset only for modded items
+	if (playerHUD->_activeItem[activeSlot].id >= CollectibleType::NUM_COLLECTIBLES) {
```
ConnorForan: Add a bound check for activeslot here just in case it is an invalid value

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1670,6 +1696,91 @@ void ASMPatchHideChargeBar() {
 	sASMPatcher.PatchAt(addr, &patch);
 }
 
+const char* __stdcall ReplaceGFXPath(const char* gfxPath, PlayerHUD* hud, int slot) {
+
+	const int collectibleID = hud->_activeItem[slot].id;
+
+	static std::string staticCustomGfxPath;
+
+	std::string customGfxPath = XMLStuff.ItemData->GetCustomActiveGFX(collectibleID);
```
ConnorForan: I think we can simplify this a bit if we modify GetCustomActiveGFX to return a `const char*` instead

IE `customActiveGFX[id].c_str()` (`return ""` will also be fine, you'll get the `const char*` to the literal)

We keep the custom active gfx strings in memory in xmldata, so it should be safe to use here unless the vanilla code were to try to delete gfxPath later, but I don't believe that happens

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1670,6 +1696,91 @@ void ASMPatchHideChargeBar() {
...
+void ASMPatchActiveItemRender() {
+	
+	void* addr = sASMDefinitionHolder->GetDefinition(&AsmDefinitions::SetActiveItemCropOffset);
+
+	printf("[REPENTOGON] Patching PlayerHUD:RenderActiveItem for capturing something at %p\n", addr);
+
+	ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::GP_REGISTERS_STACKLESS | ASMPatch::SavedRegisters::XMM_REGISTERS, true);
+	ASMPatch patch;
+	patch.PreserveRegisters(savedRegisters)
+		.Push(ASMPatch::Registers::EBP, -0x2c)
```
ConnorForan, line 1765: Generally speaking we should try to avoid using hard EBP offsets like this in patches when we can, since they can be fragile

One option is to obtain the offset via a signature for a line of code that accesses them, either based on your main signature or a different one, for example: https://github.com/TeamREPENTOGON/REPENTOGON/blob/9b290b822ac9a1c1a926ebd9f3fd8b0aa54385a0/repentogon/Patches/ASMPatches/ASMSplitTears.cpp#L204

This reads the EBP offset from a `ECX, dword ptr [EBP + ?]` (`8b 8d (90 f7 ff ff)`) (though note that sometimes the value read from the bytes is slightly offset from the value you need, it depends on the assembly command)

However, there's quite a few needed here and it can definitely be a pain to gather offsets in this way. We are also not immediately concerned with upcoming patches.

So, **I'm completely fine with you leaving the patches as they are, for now**. Just something to keep in mind for future patches :)

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1670,6 +1696,91 @@ void ASMPatchHideChargeBar() {
...
+	printf("[REPENTOGON] Patching HUD::PlayerHUD::Update for active item gfx path at %p\n", addr);
+
+	ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::GP_REGISTERS_STACKLESS & ~ASMPatch::SavedRegisters::Registers::EAX, true);
+	ASMPatch patch;
+	patch.PreserveRegisters(savedRegisters)
+		.Push(ASMPatch::Registers::EBP, -0x1024) //push activeSlot idx
+		.Push(ASMPatch::Registers::EBP, -0x1028) //push PlayerHUD
+		.Push(ASMPatch::Registers::EAX) //push original filepath to Active item sprite
+		.AddInternalCall(ReplaceGFXPath)
+		//.CopyRegister(ASMPatch::Registers::ECX, ASMPatch::Registers::EAX)
```
ConnorForan: Remove commented line if not needed, or expand comment if there's a reason for it remaining here

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1585,8 +1587,29 @@ HOOK_METHOD(PlayerHUD, RenderActiveItem, (unsigned int activeSlot, const Vector&
...
+		case COLLECTIBLE_MOVING_BOX:
+			cropOffset.x += !_player->_movingBoxContents.empty() ? 32.0f : .0f;
+			break;
+		case COLLECTIBLE_EVERYTHING_JAR:
+			/*
+			const int charges = std::max(_player->GetBatteryCharge(activeSlot), 12);
+			cropOffset.x += (float)(charges + 1) * 32.0f;
+			*/
+		case COLLECTIBLE_FLIP:
+			cropOffset.x += _player->GetPlayerType() == 38 ? 32.0f : .0f; //is dead Tainted Lazarus
```
ConnorForan: PLAYER_LAZARUS2_B (we do have this enum)

repentogon/Patches/XMLData.cpp
```diff
@@ -1189,6 +1189,11 @@ void ProcessXmlNode(xml_node<char>* node,bool force = false) {
 					} else if (id == COLLECTIBLE_DEEP_POCKETS) {
 						XMLStuff.ItemData->customcache[id].insert("maxcoins");
 					}
+
+					if (strcmp(auxnodename, "active") == 0 && item.find("activegfx") != item.end()) {
+						XMLStuff.ItemData->customActiveGFX[id] = item["gfxroot"] + "collectibles/" + item["activegfx"];
```
ConnorForan, line 1194: I was curious if we had to account for "gfxroot" not ending in a `/`, but if you do that, the normal item gfx breaks too, so we don't need to worry about that.

(No action needed, just an observation)

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
epfly6: Got it, will keep in mind

repentogon/Patches/ASMPatches/ASMCallbacks.cpp
```diff
@@ -1591,24 +1591,26 @@ HOOK_METHOD(PlayerHUD, RenderActiveItem, (unsigned int activeSlot, const Vector&
...
 		switch ((CollectibleType)activeItemID) {
 		case COLLECTIBLE_MOVING_BOX:
 			cropOffset.x += !_player->_movingBoxContents.empty() ? 32.0f : .0f;
 			break;
 		case COLLECTIBLE_EVERYTHING_JAR:
-			/*
+			
 			const int charges = std::max(_player->GetBatteryCharge(activeSlot), 12);
 			cropOffset.x += (float)(charges + 1) * 32.0f;
-			*/
```
ConnorForan: I know its still a bit pedantic since the block is commented out, but throw a `break` here

ConnorForan (approved):
lgtm, just one very minor additional comment, after which we can merge

PR #716, opened 2025-08-25
Feature/german translation and tooling

wofsauge:
Added German translation and some tools that make working with the translations a bit easier.


frto027:
oh, it already have BOM. I didn't see it clearly, please ignore the previous comment.

PR #717, opened 2025-08-27
Camera Clamping

Syntax-Sculptor:
Adds:

- Camera:SetClampEnabled(enabled)
- Camera:IsClampEnabled()

per room basis.


Syntax-Sculptor:
So this is not ready to be merged ; It turns out with the current implementation, it seems to only work with the Active Camera setting *off.*

repentogon/Patches/ASMPatches/ASMCamera.cpp
```diff
@@ -0,0 +1,54 @@
...
+bool __stdcall ShouldClampCamera() {
+	return roomCameraData.Clamped;
+}
+
+void ASMPatchCameraBoundOverride() {
+	void* clampToRoomAddr = sASMDefinitionHolder->GetDefinition(&AsmDefinitions::Camera_ClampOverride);
+
+	ASMPatch clampToRoomPatch;
+
+	clampToRoomPatch.AddInternalCall(ShouldClampCamera)
```
ConnorForan: Save registers (GP_REGISTERS_STACKLESS) before this call, and restore them AFTER your `test` (restoring stack registers does not overwrite the test bits)

repentogon/Patches/ASMPatches/ASMCamera.cpp
```diff
@@ -0,0 +1,54 @@
...
+		.AddBytes("\x84\xC0") // test al, al
+		.AddConditionalRelativeJump(ASMPatcher::CondJumps::JNZ, (char*)clampToRoomAddr + 0x14)
+		.AddBytes("\x8B\x45\x08") // mov eax, dword ptr [ebp + 0x8]
+		.AddRelativeJump((char*)clampToRoomAddr + 0x5);
+
+	sASMPatcher.PatchAt(clampToRoomAddr, &clampToRoomPatch);
+
+	ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::ALL & ~ASMPatch::SavedRegisters::Registers::EDX, false);
```
ConnorForan: ALL -> GP_REGISTERS_STACKLESS

repentogon/Patches/ASMPatches/ASMCamera.cpp
```diff
@@ -0,0 +1,54 @@
...
+	void* ultraSmoothAddr = sASMDefinitionHolder->GetDefinition(&AsmDefinitions::Camera_SlowStopClampOverride);
+
+	ASMPatch smoothUpdatePatch;
+	smoothUpdatePatch.Push(ASMPatch::Registers::EDX)
+		.PreserveRegisters(savedRegisters)
+		.AddInternalCall(ShouldClampCamera)       
+		.AddBytes("\x0F\xB6\xD0") // movzx edx, al
+		.RestoreRegisters(savedRegisters)
+		.AddBytes("\x85\xD2") // test edx, edx
```
ConnorForan: Similarly to the other patch, you should be able to `test` using eax after calling your function, THEN restore registers, THEN do your conditional jump. That way you should be able to skip all the edx stuff

1. Preserve
2. Call
3. test eax, eax
4. Restore
5. conditional jump

repentogon/Patches/ASMPatches/ASMCamera.cpp
```diff
@@ -0,0 +1,56 @@
...
+	ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::GP_REGISTERS_STACKLESS, true);
+
+	ASMPatch clampToRoomPatch;
+
+	clampToRoomPatch.PreserveRegisters(savedRegisters)
+		.AddInternalCall(ShouldClampCamera)
+		.AddBytes("\x84\xC0") // test al, al
+		.RestoreRegisters(savedRegisters)
+		.AddConditionalRelativeJump(ASMPatcher::CondJumps::JNZ, (char*)clampToRoomAddr + 0x14)
+		.AddBytes("\x8B\x45\x08") // mov eax, dword ptr [ebp + 0x8]
```
ConnorForan: Since this seems to be a recreation of overwritten bytes, would be nice to explicitly copy them here (safer against offset changes)

probably something like

`.AddBytes(ByteBuffer().AddAny((char*)clampToRoomAddr + 0x2, 0x3))`

(copies the 3 bytes after the jump that this patch starts at, so 8b 45 08)

ConnorForan (approved):
One more small comment but otherwise the patches look fine

PR #733, opened 2025-09-25
Native Custom Curse Icon support and extension to ModEntry

epfly6:
This PR introduces custom icons support for modded curses (and extended version of ModManager/ModEntry development for potential development of similar features on C++)
image

Custom graphics can be defined in mapitemicons.anm2, placed in ``modName/content/gfx`` folder, with one frame per curse entry on animation `curses`, otherwise all icons from the mod will be empty by default.

Recommended settings for curse frame is 16x16 and centered around the origin (position set to 0, 0, pivot to 8, 8)
image

Sample mod included: 
[SampleCustomCurseIcons.zip](https://github.com/user-attachments/files/22862592/SampleCustomCurseIcons.zip)

### DIFF
--- a/repentogon/Patches/CustomModManager.cpp
+++ b/repentogon/Patches/CustomModManager.cpp
@@ -129,7 +129,7 @@ void ASMPatchAssignCustomFrame() {
 	printf("[REPENTOGON] Patching Minimap::render_icons for accessing mapIcons array at %p\n", addr);
 
 	patch.PreserveRegisters(savedRegisters)
-		.LoadEffectiveAddress(ASMPatch::Registers::ESP, mapIconsOffset + 0x14, ASMPatch::Registers::EDI, std::nullopt, 4u ) //load mapIcons array ptr to EDI
+		.LoadEffectiveAddress(ASMPatch::Registers::ESP, mapIconsOffset + 0x14, ASMPatch::Registers::EDI) //load mapIcons array ptr to EDI
 		.Push(ASMPatch::Registers::EBX) // push iconCount
 		.Push(ASMPatch::Registers::EDI) //push mapIcons array
 		.Push(ASMPatch::Registers::EAX) //push curseBitmask
@@ -176,9 +176,9 @@ void ASMPatchRenderCustomCurses() {
 	printf("[REPENTOGON] Patching Minimap::render_icons for rendering custom curses at %p\n", addr);
 
 	patch.PreserveRegisters(savedRegisters)
-		.LoadEffectiveAddress(ASMPatch::Registers::ESP, colorModAlphaOffset + 0x20, ASMPatch::Registers::EBX, std::nullopt, 4u) // load color alpha ptr to EBX
+		.LoadEffectiveAddress(ASMPatch::Registers::ESP, colorModAlphaOffset + 0x20, ASMPatch::Registers::EBX) // load color alpha ptr to EBX
 		.Push(ASMPatch::Registers::EBX)
-		.LoadEffectiveAddress(ASMPatch::Registers::ESP, positionOffset + 0x14, ASMPatch::Registers::EAX, std::nullopt, 4u) // load position ptr to EAX
+		.LoadEffectiveAddress(ASMPatch::Registers::ESP, positionOffset + 0x14, ASMPatch::Registers::EAX) // load position ptr to EAX
 		.Push(ASMPatch::Registers::EAX) //push position
 		.Push(ASMPatch::Registers::EDI) // push curseId
 		.AddInternalCall(RenderCustomCurses)


repentogon/Patches/CustomModManager.h
```diff
@@ -0,0 +1,37 @@
...
+        auto it = modExtensions.find(mod);
+        return (it != modExtensions.end()) ? &it->second : nullptr;
+    }
+
+    void AddEx(ModEntry* mod) {
+        modExtensions[mod];
+    }
+
+private:
+    std::map<ModEntry*, ModEntryEx> modExtensions;
```
ConnorForan: May as well use std::unordered_map

repentogon/Patches/CustomModManager.cpp
```diff
@@ -0,0 +1,227 @@
...
+void __stdcall RegisterCurseSprite(uint32_t curseId, bool isMod) {
+	if (!isMod || !capturedModEntry) return;  // No mod captured
+
+	if (capturedModEntry != lastModEntry) {
+		currentFrameNum = 0;
+		lastModEntry = capturedModEntry;
+	}
+
+	ModEntryEx* ex = CustomModManager::GetInstance().GetEx(capturedModEntry);
+	if (ex && &ex->_customMinimapANM2) {
```
ConnorForan: It's not possible for `&ex->_customMinimapANM2` to be nullptr since `_customMinimapANM2` is a whole field in ModEntryEx

However, you should maybe check `ex->_customMinimapANM2._loaded` ?

repentogon/Patches/CustomModManager.cpp
```diff
@@ -0,0 +1,227 @@
...
+		.AddInternalCall(CaptureModEntry)
+		.RestoreRegisters(savedRegisters)
+		.AddBytes(ByteBuffer().AddAny((char*)addr, 0x6))  // Restore the bytes we overwrote
+		.AddRelativeJump((char*)addr + 0x6);
+	sASMPatcher.PatchAt(addr, &patch);
+}
+
+HOOK_METHOD(RoomConfig, LoadCurses, (char* xmlPath, bool isMod) -> void) {
+	if (isMod) {
+		LoadCustomMinimapANM2(capturedModEntry);
```
ConnorForan, line 64: I was originally going to ask why this is called both here and in ASMPatchLoadModEntryExAssets but I see that patch is disabled.

I guess remove or comment out the patch if its not needed, if loading the anm2s in the LoadCurses hook works that feels ideal

repentogon/Patches/CustomModManager.cpp
```diff
@@ -0,0 +1,227 @@
...
+		.Push(ASMPatch::Registers::EBP, 0xC) //isMod
+		.Push(ASMPatch::Registers::EBP, curseIdOffset)
+		.AddInternalCall(RegisterCurseSprite)
+		.RestoreRegisters(savedRegisters)
+		.AddBytes(ByteBuffer().AddAny((char*)addr, 0xA))  // Restore the bytes we overwrote
+		.AddRelativeJump((char*)addr + 0xA);
+	sASMPatcher.PatchAt(addr, &patch);
+}
+
+int __stdcall TestArrayPrint(int* mapIcons, int mapIndex) {
```
ConnorForan: test thing left in

repentogon/Patches/CustomModManager.cpp
```diff
@@ -0,0 +1,227 @@
...
+	ASMPatch patch;
+
+	const int8_t mapIconsOffset = *(int8_t*)((char*)sASMDefinitionHolder->GetDefinition(&AsmDefinitions::CustomModManager_CurseMapIconsArrayOffset) + 0x3);
+	void* addr = sASMDefinitionHolder->GetDefinition(&AsmDefinitions::CustomModManager_CurseRegisterFrameOffset);
+	const int jumpOffset = 0x7 + *(int8_t*)((char*)addr + 0x6);
+
+	printf("[REPENTOGON] Patching Minimap::render_icons for accessing mapIcons array at %p\n", addr);
+
+	patch.PreserveRegisters(savedRegisters)
+		.LoadEffectiveAddress(ASMPatch::Registers::ESP, mapIconsOffset + 0x14, ASMPatch::Registers::EDI, std::nullopt, 4u ) //load mapIcons array ptr to EDI
```
ConnorForan: If I understand correctly, you can omit the `std::nullopt, 4u` from the LEA calls entirely, and that should be fine?

IE, `.LoadEffectiveAddress(ASMPatch::Registers::ESP, mapIconsOffset + 0x14, ASMPatch::Registers::EDI)`

the 4u is the "scale" value, which is multiplied with the "index" register, the one you have set to nullopt, so I think it does nothing here

If I've misunderstood something and that doesn't work, my bad

https://github.com/TeamREPENTOGON/REPENTOGON/blob/43c4350eaabac87c3ebab1e2a84853936c81f5b7/libzhl/ASMPatcher.hpp#L416

https://github.com/TeamREPENTOGON/REPENTOGON/blob/43c4350eaabac87c3ebab1e2a84853936c81f5b7/libzhl/ASMPatcher.hpp#L204

repentogon/Patches/CustomModManager.h
epfly6: Added description of PR and sample mod

repentogon/Patches/CustomModManager.cpp
epfly6: Mod will terminate if base Register is ESP and scale is set to 0. See: https://github.com/TeamREPENTOGON/REPENTOGON/blob/43c4350eaabac87c3ebab1e2a84853936c81f5b7/libzhl/ASMPatcher.cpp#L817-L838

repentogon/Patches/CustomModManager.h
epfly6: Changed type

repentogon/Patches/CustomModManager.cpp
epfly6: Swapped check to ex->_customMinimapANM2._loaded

repentogon/Patches/CustomModManager.cpp
epfly6: Removed patch that used LoadCustomMinimapANM2, since ModManager::UpdateCurses triggers before ModManager::LoadConfigs

repentogon/Patches/CustomModManager.cpp
epfly6: Axed to horrible death

repentogon/Patches/CustomModManager.cpp
Sylmir: Woops, kinda forgot this was a thing, need to be fixed on my side

Sylmir: Hopefully fixed in https://github.com/TeamREPENTOGON/REPENTOGON/commit/32c265b27ed693d6182f2f6c4489694cd93de431

PR #744, opened 2025-10-15
Fix issues when rendering Images with any Shader

Guantol-Lemat:
This PR aims to fix issues in the KAGE engine that prevented an Image from being rendered when using an "incompatible" Shader type. The root cause being heap corruption that occurred when attempting to batch render images whose shaders had mismatched `VertexDescriptors`.

Because the batching process bases vertex metadata on the Image structure - rather than on the Shader itself - it is still not possible to directly use the Image methods to, safely, render images. As such a utility method has been provided that temporarily modifies the Image data, so that batching can proceed without issues.

Since the ANM2 render function's pool of shaders are not compatible with each other (ColorOffsetChampion is incompatible with all the other) all calls to `Image:Render` had to be replaced to use the Utility function. This was achieved by **Reimplementing** the `AnimationLayer::RenderFrame` function.

The reimplementation was originally created as a temporary override for debugging render behavior, but the function proved to be simple to recreate, in part due to some of the sub-operation not being inlined. This, alongside the improved performance it offers to all render operations, and in view of allowing fully custom shaders to be used when rendering; I decided to keep it, and not switch to an ASM patch.

### DIFF
--- a/repentogon/Utils/ShaderUtils.hpp
+++ b/repentogon/Utils/ShaderUtils.hpp
@@ -1,3 +1,5 @@
+#pragma once
+
 #include "ShaderUtils.h"
 
 inline size_t ShaderUtils::GetNumVertexAttributes(KAGE_Graphics_ImageBase_VertexAttributeDescriptor *descriptor) noexcept


repentogon/Utils/ANM2Utils.cpp
ConnorForan, line 1: Minor, but I gotta ask... what's the deal with having all three of .h + .hpp + .cpp, with the last one being basically empty

ConnorForan (approved):
LGTM. Just a couple minor questions.

Thanks for including some details on why RenderFrame was **Reimplemented**. No, I have no objections with it... the function is not overly large or complex, especially with the Quads and Vertices-related functionality factored out, which appears to be things we needed for the fixes in this PR as well as upcoming features. It makes enough sense given what you want to accomplish, and does open up some options for us.

One thing I am curious about, you mention improved performance for RenderFrame, so I was just curious what you think the reason for that might be. Is it due to the ImagePatches, or the new Quads/Vertices functions? From some very simple benchmarking it did indeed seem to be very slightly faster.

Guantol-Lemat:
The only thing I improve from the original function is not calling the virtual getters for Width and Height multiple times during FillVertices, and only calling GetSpriteSheet once for the duration of the function, whilst the original would call it every time the image was needed, and since that involves a lazy sync + a SmartPointer copy and subsequent SmartPointer destruction it might start to pile some slight overhead for each of the rendered frames.

However I suspect the biggest contributor might be that I simply did not call `super` inside of the hook, like the previous implementation of CustomShaders did. I noticed that when I accidentally set the hook to have high priority instead of low priority, the function turned out to be about 2x faster, despite the other hooks being quite simple and not as heavy, computational wise, compared to the whole RenderFrame function (InterpolationV2 was disabled on my end so it amounted to an if check, and LuaRender's scoped context constructor and destructor seems to just be a deque push_back, and a pop_back respectively).

repentogon/Utils/ANM2Utils.cpp
Guantol-Lemat: It was something I did so that users of the functions could choose whether to just include the .h file to avoid recompiling utility implementations, or include the .hpp to expose inline definitions and let the compiler optimize/inline calls.

Since the utils contain few functions now, including the .hpp will probably be better and not impact compilation times, but I still wanted to give this option.

PR #750, opened 2025-11-03
Custom level generation

thicco-catto:
Added a new callback and a couple functions that allow for mods to generate custom floors.

**MC_PRE_GENERATE_DUNGEON**
Called before `level::generate_dungeon` is called, which places all the rooms in a floor. Passes an RNG.
Return true to cancel vanilla floor generation.

**Level:ResetRoomList(bool)**
Needs to be called before placing any rooms in the custom level.
The bool is the same as the one in `level::init`. Testing it so far, setting it to true makes it work all the time.

**Level:SetLastBossRoomListIndex(int)**
Also needs to be called after creating the custom floor, since not setting it crashes the game when continuing.
Doesn't actually need to be set to a boss room for the game to not crash, but I haven't tested any other side effects apart from boss rooms not spawning the trapdoor if they are not the last. 

**Sample**
image
```lua
local TestMod = RegisterMod("Test Mod", 1)

local function place_at(room, col, row, seed)
    local level = Game():GetLevel()

    local entry = Isaac.LevelGeneratorEntry()
    entry:SetAllowedDoors(15)
    entry:SetColIdx(col)
    entry:SetLineIdx(row)

    level:PlaceRoom(entry, room, seed)
end

local function get_normal_room(seed)
    return RoomConfigHolder.GetRandomRoom(seed, false, StbType.BASEMENT, RoomType.ROOM_DEFAULT, RoomShape.ROOMSHAPE_1x1, -1, -1, 0, 10, 15)
end

TestMod:AddCallback(ModCallbacks.MC_PRE_GENERATE_DUNGEON, function (_, rng)
    local level = Game():GetLevel()
    level:ResetRoomList(true)

    print(rng:GetSeed())

    local start_room = RoomConfigHolder.GetRandomRoom(rng:Next(), false, StbType.SPECIAL_ROOMS, RoomType.ROOM_DEFAULT, RoomShape.ROOMSHAPE_1x1, 2, 2)
    local big_room = RoomConfigHolder.GetRandomRoom(rng:Next(), false, StbType.BASEMENT, RoomType.ROOM_DEFAULT, RoomShape.ROOMSHAPE_2x2, -1, -1, 0, 10, 255)
    local boss_room = RoomConfigHolder.GetRandomRoom(rng:Next(), false, 0, RoomType.ROOM_BOSS, RoomShape.ROOMSHAPE_1x1)

    place_at(start_room, 6, 6, rng:Next())
    place_at(get_normal_room(rng:Next()), 7, 6, rng:Next())
    place_at(boss_room, 8, 6, rng:Next())
    place_at(get_normal_room(rng:Next()), 6, 7, rng:Next())
    place_at(get_normal_room(rng:Next()), 6, 8, rng:Next())
    place_at(get_normal_room(rng:Next()), 7, 8, rng:Next())
    place_at(get_normal_room(rng:Next()), 8, 7, rng:Next())
    place_at(big_room, 8, 8, rng:Next())

    level:SetLastBossRoomListIndex(2)

    return true
end)
```


repentogon/Patches/ASMPatches/ASMLevel.cpp
ConnorForan, line 1: The code added here is actually in the wrong place, technically. This directory is primarily for Assembly (ASM) Patches

Since this is just function hooks, should probably move this stuff to `repentogon/LuaInterfaces/CustomCallbacks.cpp`

repentogon/Patches/ASMPatches/ASMLevel.cpp
```diff
@@ -82,8 +83,67 @@ void ASMPatchVoidGeneration() {
...
+		.push(generator, lua::metatables::DungeonGeneratorMT)
+		.push(rng, lua::Metatables::RNG)
+		.call(1);
+
+	if (results || !lua_isboolean(L, -1) || !lua_toboolean(L, -1))
+	{
+		return false;
+	}
+
+	level->reset_room_list(false);
```
ConnorForan: I feel like it might make sense for the logic to reset & build out the floor be a DungeonGenerator function (like `generator->Generate()`)

May also be good for it to return a success boolean, in case we decide it failed somehow, we could maybe fall back to allowing the vanilla generation to happen again.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "Exception.h"
+#include "Log.h"
+#include "LuaDungeonGenerator.h"
+
+DungeonGenerator* GetDungeonGenerator(lua_State* L) {
+	return *lua::GetRawUserdata<DungeonGenerator**>(L, 1, lua::metatables::DungeonGeneratorMT);
+}
+
+LUA_FUNCTION(place_room) {
+	DungeonGenerator* generator = GetDungeonGenerator(L);
```
ConnorForan, line 596: I think it would be good for DungeonGenerator itself to contain all its functions such as PlaceRoom (ie, `generator->PlaceRoom(config, row, col, seed)`, and the lua bindings just parse the inputs from lua and redirect to the generator's functions. This will separate the DungeonGenerator logic from the lua bindings, and allow DungeonGenerator code to call its own functions.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "LuaCore.h"
+#include "Exception.h"
+#include "Log.h"
+#include "LuaDungeonGenerator.h"
+
+DungeonGenerator* GetDungeonGenerator(lua_State* L) {
+	return *lua::GetRawUserdata<DungeonGenerator**>(L, 1, lua::metatables::DungeonGeneratorMT);
+}
+
+LUA_FUNCTION(place_room) {
```
ConnorForan: nitpick: Use names like `Lua_PlaceRoom` for the lua bindings, mostly just a consistency thing (not that our consistency is fantastic overall)

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
+#include "HookSystem.h"
+#include "IsaacRepentance.h"
+#include "LuaCore.h"
+#include "Exception.h"
+#include "Log.h"
+#include "LuaDungeonGenerator.h"
+
+DungeonGenerator* GetDungeonGenerator(lua_State* L) {
```
ConnorForan: `LUALIB_API DungeonGenerator* GetDungeonGenerator(lua_State* L)`

I don't know exactly how important the `LUALIB_API` bit is, but its typically used for such functions. Might impact properly surfacing lua errors.

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,34 @@
...
+		this->seed = seed;
+	}
+};
+
+struct DungeonGenerator {
+	int a;
+	int num_rooms;
+	DungeonGeneratorRoom rooms[169];
+
+	DungeonGenerator() {
```
ConnorForan: Might be nice if we initialize DungeonGenerator using an RNG* (ie, the one the game provided)

That way it won't be necessary for the user to specify seeds, and allows the DungeonGenerator to do seeded rng at will

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
...
+				generator_room->is_final_boss = true;
+			} else {
+				generator_room->is_final_boss = false;
+			}
+		} else {
+			break;
+		}
+	}
+
+	return 1;
```
ConnorForan, line 819: For lua bindings, `return 1` means that the lua function returned something back to lua. Specifically, I believe the number is how many things were pushed to the lua stack.

If we dont return anything, we should `return 0`

Though, if we add any validation, this function could maybe return true on success (`lua_pushboolean(L, value);`)

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
...
+	generator->rooms[generator->num_rooms] = *generatorRoom;
+
+	generator->num_rooms++;
+
+	return 1;
+}
+
+LUA_FUNCTION(set_final_boss_room) {
+	DungeonGenerator* generator = GetDungeonGenerator(L);
+	uint32_t row = (uint32_t)luaL_checkinteger(L, 2);
```
ConnorForan: Suggested in external discussion: Take a DungeonGeneratorRoom instead of column+row

Should we enforce that this is only used on a room with a Boss Room config?

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
...
+	uint32_t row = (uint32_t)luaL_checkinteger(L, 3);
+	uint32_t col = (uint32_t)luaL_checkinteger(L, 4);
+	uint32_t seed = (uint32_t)luaL_checkinteger(L, 5);
+
+	DungeonGeneratorRoom* generatorRoom = new DungeonGeneratorRoom(config, row, col, seed);
+	generator->rooms[generator->num_rooms] = *generatorRoom;
+
+	generator->num_rooms++;
+
+	return 1;
```
ConnorForan, line 695: Suggested in external discussion: Return DungeonGeneratorRoom

Example:
```
DungeonGeneratorRoom** ud = (DungeonGeneratorRoom**)lua_newuserdata(L, sizeof(DungeonGeneratorRoom*));
*ud = generatorRoom;
luaL_setmetatable(L, lua::metatables::DungeonGeneratorRoomMT);

return 1;
```

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,34 @@
...
+	DungeonGeneratorRoom(RoomConfig_Room* room, uint32_t row, uint32_t col, uint32_t seed) {
+		this->room = room;
+		this->row = row;
+		this->col = col;
+		this->seed = seed;
+	}
+};
+
+struct DungeonGenerator {
+	int a;
```
ConnorForan: What is `a`?

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,34 @@
...
+
+struct DungeonGeneratorRoom {
+	RoomConfig_Room* room;
+	uint32_t row;
+	uint32_t col;
+	uint32_t seed;
+	bool is_final_boss = false;
+	
+	DungeonGeneratorRoom() {
+		this->room = NULL;
```
ConnorForan: nit: Prefer `nullptr` instead of `NULL`

repentogon/Patches/ASMPatches/ASMLevel.cpp
```diff
@@ -82,8 +83,67 @@ void ASMPatchVoidGeneration() {
...
+	}
+
+	g_Game->_nbRooms = 0;
+
+	for (size_t i = 0; i < generator->num_rooms; i++)
+	{
+		DungeonGeneratorRoom generator_room = generator->rooms[i];
+
+		if (generator_room.room != NULL) {
+			LevelGenerator_Room* level_generator_room = new LevelGenerator_Room();
```
ConnorForan: As pointed out by Guantol, this is a memory leak atm

You can initialize a local variable class/struct as just `LevelGenerator_Room level_generator_room;` and it will be cleaned up when it leaves scope as you'd expect. This also calls the default constructor (for classes at least) automatically where applicable.

Initializing classes with `new` requires explicit freeing later, though there's usually better options than `new`

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+LUA_FUNCTION(place_room) {
+	DungeonGenerator* generator = GetDungeonGenerator(L);
+	RoomConfig_Room* config = lua::GetLuabridgeUserdata<RoomConfig_Room*>(L, 2, lua::Metatables::CONST_ROOM_CONFIG_ROOM, "RoomConfig");
+	uint32_t row = (uint32_t)luaL_checkinteger(L, 3);
+	uint32_t col = (uint32_t)luaL_checkinteger(L, 4);
+	uint32_t seed = (uint32_t)luaL_checkinteger(L, 5);
+
+	DungeonGeneratorRoom* generatorRoom = new DungeonGeneratorRoom(config, row, col, seed);
```
ConnorForan: This initializes a DungeonGeneratorRoom, and then saves a COPY of it into the array on the next line. The original is also a memory leak.

I think you can initialize it like this:

```
generator->rooms[generator->num_rooms] = DungeonGeneratorRoom(config, row, col seed);
generator->num_rooms++;
DungeonGeneratorRoom* generatorRoom = &generator->rooms[generator->num_rooms];
```

This should initialize a DungeonGeneratorRoom inline and assigns it into the array (no copying and no memory management required).

The three example line above would probably be good as a small helper function (ie, `DungeonGeneratorRoom* generatorRoom = generator->CreateRoom();`?

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+DungeonGenerator* GetDungeonGenerator(lua_State* L) {
+	return *lua::GetRawUserdata<DungeonGenerator**>(L, 1, lua::metatables::DungeonGeneratorMT);
+}
+
+LUA_FUNCTION(place_room) {
+	DungeonGenerator* generator = GetDungeonGenerator(L);
+	RoomConfig_Room* config = lua::GetLuabridgeUserdata<RoomConfig_Room*>(L, 2, lua::Metatables::CONST_ROOM_CONFIG_ROOM, "RoomConfig");
+	uint32_t row = (uint32_t)luaL_checkinteger(L, 3);
+	uint32_t col = (uint32_t)luaL_checkinteger(L, 4);
```
ConnorForan: Validate the row/col (both bounds checking and verifying that there isn't already a room in that slot)

In theory we'll also need to do other stuff like handling overlapping large rooms and checking door slots, but you don't HAVE to add that right now, unless you want to

repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -5646,4 +5647,90 @@ HOOK_STATIC(LuaEngine, PostEntityKill, (Entity* ent) -> void, __stdcall) {
...
+	}
+
+	bool correctGeneration = generator.Generate(level);
+
+	return correctGeneration;
+}
+
+HOOK_METHOD(Level, generate_dungeon, (RNG* rng) -> void)
+{
+	bool skip = ProcessGenerateDungeonCallback(this, *rng, 0);
```
Guantol-Lemat: Minor nitpick, but I would suggest replacing the magic numbers for the DungeonType into either an enum or a constexpr, to make the code more readable.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,578 @@
...
+	std::vector<RoomCoords> occupied_coords = GetOccupiedCoords(base_coords, room_shape);
+	for (RoomCoords coords : occupied_coords) {
+		if (!coords.IsValid()) {
+			return false;
+		}
+	}
+
+	std::vector<RoomCoords> forbidden_coords = GetForbiddenNeighbors(base_coords, room_shape, doors);
+
+	for (int i = 0; i < this->num_rooms; i++) {
```
Guantol-Lemat: I would suggest having the generator hold a map that ties gridIdx to roomIdx, in order to quickly find the neighboring rooms, and see if a room already occupies this room's slot, as going through the whole room list is not really efficient.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+DungeonGeneratorRoom* DungeonGenerator::PlaceRoom(RoomConfig_Room* room_config, uint32_t col, uint32_t row, int doors) {
+	this->rooms[this->num_rooms] = DungeonGeneratorRoom(room_config, col, row, doors);
+	this->num_rooms++;
+	DungeonGeneratorRoom* generatorRoom = &this->rooms[this->num_rooms - 1];
+
+	return generatorRoom;
+}
+
+void DungeonGenerator::SetFinalBossRoom(DungeonGeneratorRoom* boss_room) {
```
Guantol-Lemat, line 453: I would have the generator hold a single value that defines the index of the final boss room. This would trivially guarantee the constraint that only one room can be the FinalBossRoom, as well as being faster to set and check.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,578 @@
...
+	}
+
+	bool operator==(RoomCoords const& other) const {
+		return this->row == other.row && this->col == other.col;
+	}
+};
+
+#pragma endregion
+
+#pragma region Helpers
```
Guantol-Lemat: I would suggest placing these functions inside of their own Utils file, since they can also reused internally and are not strictly for the LuaDungeonGenerator.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,578 @@
...
+		return 1;
+	} else {
+		return 0;
+	}
+}
+
+LUA_FUNCTION(Lua_PlaceDefaultStartingRoom) {
+	DungeonGenerator* generator = GetDungeonGenerator(L);
+
+	int doors = (int)luaL_optinteger(L, 2, 15);
```
Guantol-Lemat, line 806: Since the start room has essentially a predefined config, and GetRandomRoom can only ever pick one room, due to the variant range being set to just the value 2, there is no need to pass the doors parameter.

repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -5646,4 +5647,90 @@ HOOK_STATIC(LuaEngine, PostEntityKill, (Entity* ent) -> void, __stdcall) {
...
+	if (skip) {
+		return;
+	}
+
+	super(rng);
+}
+
+HOOK_METHOD(Level, generate_blue_womb, () -> void) {
+	bool skip = ProcessGenerateDungeonCallback(this, g_Game->_generationRNG, 1);
```
Guantol-Lemat: For now I would suggest forcibly making it impossible to generate layouts on dungeon types other than NORMAL, perhaps with an error message explaining that they are currently not supported, since we need to properly evaluate the constraints needed to make the other dungeon types work properly first.

libzhl/functions/Level.zhl
```diff
@@ -7,6 +7,9 @@ __thiscall void Level::Init(bool unkinitStartRoom);
 "518b89????????e8":
 __thiscall void Level::Update();
 
+"558bec6aff68????????64a1????????5081ecc0000000a1????????33c58945??535657508d45??64a3????????8bc1":
+__thiscall void Level::reset_room_list(bool unkinitStartRoom);
```
Guantol-Lemat: minor nitpick, change the name of the bool parameter to 'resetLilPortalRoom'

repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -5646,4 +5647,90 @@ HOOK_STATIC(LuaEngine, PostEntityKill, (Entity* ent) -> void, __stdcall) {
...
+	DungeonGenerator generator(&rng);
+	lua::LuaResults results = lua::LuaCaller(L)
+		.push(callbackId)
+		.push(dungeonType)
+		.push(&generator, lua::metatables::DungeonGeneratorMT)
+		.push(&rng, lua::Metatables::RNG)
+		.push(dungeonType)
+		.call(1);
+
+	if (results || !lua_isboolean(L, -1) || !lua_toboolean(L, -1))
```
Guantol-Lemat: We might want to change this, as right now I expect that if someone returns an invalid level layout, we don't give a chance for other callbacks to run and simply return to default generation.

repentogon/Patches/ASMPatches/ASMLevel.cpp
```diff
@@ -5,6 +5,7 @@
 #include "ASMLevel.h"
 #include "../../LuaInterfaces/Level.h"
 #include "../../LuaInterfaces/Room/Room.h"
+#include "../../LuaInterfaces/LuaDungeonGenerator.h"
```
Guantol-Lemat: From my understanding this include is unused, so it should be removed.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
thicco-catto: The doors parameter limits the allowed doors for the room. It could be useful to mimick the behaviour of the mega satan door/polaroid door

repentogon/LuaInterfaces/CustomCallbacks.cpp
thicco-catto: Would it be better to just skip the callback call and write a note in the docs explaining that the callback only works for default generation?

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
Guantol-Lemat: It would be better to use a system similar to the blocked grid indices, rather than having the user having to deal with doors directly. In the original level gen the doors parameter is meant as a way to specify which doors are **necessary**, not those that are allowed. Of course, we should still allow control over the doors parameter, when doing very specific things, like how secret rooms and ultra secret room generation has to mark certain doors as allowed, post layout generation.

thicco-catto: I was basing the doors parameter on the LevelGeneratorEntry.doors parameter, since I'm later setting the LevelGeneratorEntry doors field to that. There it does specify the doors argument refers to the allowed doors.
https://repentogon.com/LevelGeneratorEntry.html

Guantol-Lemat: LevelGeneratorEntry is a wrapper for a LevelGenerator_Room and LevelGenerator_Room only specifies the necessary rooms, not the allowed ones. Most likely the parameter was named like that because the necessary rooms end up being used as the allowed rooms when placing the room.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -333,91 +31,88 @@ DungeonGeneratorRoom::DungeonGeneratorRoom(RoomConfig_Room* room, uint32_t col,
 	this->shape = room->Shape;
 }
 
+RoomConfig_Room* DungeonGeneratorRoom::GetRoomConfig(uint32_t seed, int required_doors) {
```
Guantol-Lemat: I presume this function was originally meant to do something else due to the unused parameters and unnecessary check for nullptr, what was this supposed to do originally?

repentogon/LuaInterfaces/CustomCallbacks.cpp
ConnorForan: The logic here seems like it might be fine, but yeah this callback will likely require custom handling in main_ex.lua. We could potentially detect that a mod left things in an invalid state, possibly print an error to the console, then reset it before we run the next callback.

Not mandatory for an initial push, but is something we should sort out before this gets included in a release.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+	this->level_generator._rng = *rng;
+	std::fill_n(this->level_generator._roomMap, 169, -1);
+	std::fill_n(this->level_generator._blockedPositions, 169, false);
+	this->level_generator._isChapter6 = false;
+	this->level_generator._isStageVoid = false;
+	this->level_generator._isXL = false;
+}
+
+bool DungeonGenerator::CanRoomBePlaced(XY& base_coords, int shape, int allowed_doors, bool allow_unconnected) {
```
ConnorForan, line 115: When placing rooms with RoomConfig_Rooms, we may need to validate somewhere that the two rooms are actually able to mutually connect allowed doorslots to each other. I dont think vanilla really does this since iirc it specifically queries rooms with specific doors available when filling the layout with configs

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,278 @@
...
+			generator->rng->Next(),
+			false,
+			STB_SPECIAL_ROOMS,
+			ROOM_DEFAULT,
+			ROOMSHAPE_1x1,
+			2,
+			2,
+			0,
+			10,
+			(unsigned int*)&required_doors, // If I don't do it like this it shits itself
```
ConnorForan: Just make the `int required_doors` variable an `unsigned int` to begin with so you dont have to do the weird casting

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,278 @@
...
+		DungeonGeneratorRoom* generator_room = generator->PlaceRoom(config, col, row, doors);
+
+		DungeonGeneratorRoom** ud = (DungeonGeneratorRoom**)lua_newuserdata(L, sizeof(DungeonGeneratorRoom*));
+		*ud = generator_room;
+		luaL_setmetatable(L, lua::metatables::DungeonGeneratorRoomMT);
+
+		return 1;
+	}
+	else {
+		return 0;
```
ConnorForan: In cases like this, instead of returning 0:
```
lua_pushnil(L);
return 1;
```
Prefer to only return 0 for functions that never return anything.

Dunno how severely this matters, but it seems like the consistent thing to do

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
thicco-catto: This is already handled by the allowed_doors parameter. When a room is placed in the layout it blocks all inaccessible grid indexes

repentogon/LuaInterfaces/CustomCallbacks.cpp
ConnorForan: Handling was added to main_ex.lua. Resolving.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
ConnorForan: bump.

Remove unused parameter, and simply `return this->room;`

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
ConnorForan: Seems like this change was made, so resolving.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
ConnorForan: Bump.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
ConnorForan: Bump. Refactor this logic as DungeonGenerator::PlaceRoom, and just call that from here.

IE, make lua functions minimal (get params from lua, push return values) and do most other things within the DungeonGenerator

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
ConnorForan: Bump. IDK if I have input here but need to decide how we want to handle non-standard door layouts for the starting room.

Maybe DM guantol.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,344 @@
...
+	if (generator->CanRoomBePlaced(coords, config->Shape, allowed_doors, true)) {
+		DungeonGeneratorRoom* generator_room = generator->PlaceRoom(config, col, row, allowed_doors);
+
+		DungeonGeneratorRoom** ud = (DungeonGeneratorRoom**)lua_newuserdata(L, sizeof(DungeonGeneratorRoom*));
+		*ud = generator_room;
+		luaL_setmetatable(L, lua::metatables::DungeonGeneratorRoomMT);
+
+		return 1;
+	} else {
+		return 0;
```
ConnorForan: I'd prefer this
```c++
  ...
  luaL_setmetatable(L, lua::metatables::DungeonGeneratorRoomMT);
} else {
  lua_pushnil(L);
}
return 1;
```

repentogon/resources/scripts/main_ex.lua
```diff
@@ -1687,6 +1687,21 @@ local function RunPreStatusEffectApplyCallback(callbackID, param, status, entity
 	return recentRet
 end
 
+local function RunPreGenerateDungeonCallback(callbackID, param, dungeonGenerator, rng, dungeonType)
+	for callback in GetCallbackIterator(callbackID, param) do
+		local ret = RunCallbackInternal(callbackID, callback, dungeonGenerator, rng, dungeonType)
+
+		if type(ret) == "boolean" and ret then
```
ConnorForan: I think this is very little room for any semblance of mod compatibility for this callback. I believe we've discussed this a bit before.

MC_PRE_GENERATE_DUNGEON is simply not suitable for modifying layouts, only creating them. The use-case being served here is completely custom layouts, and that is not something that mods can collaborate on without explicit compatibility.

With that in mind, I wonder if instead of having a return value at all, we could just do this:
```lua
RunCallbackInternal(callbackID, callback, dungeonGenerator, rng, dungeonType)
if dungeonGenerator:Validate() then
  return true
else
  dungeonGenerator:Reset()
end
```

If there was a version of this that was intended to support modifications, it would either be `MC_POST_GENERATE_DUNGEON`, or a more fleshed out `MC_POST_LEVEL_LAYOUT_GENERATED`. But that is a weird case. Tacking additional rooms onto the floor is easy enough to do now post-levelgen, so this would come down to practical use-cases for modifying layouts that cannot simply be done later. Stuff to think about/discuss separately, perhaps.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,344 @@
...
+			return false;
+		}
+
+		uint32_t seed = this->rng->Next();
+
+		g_Game->PlaceRoom(&room, room_config, seed, 0);
+	}
+
+	g_Game->_lastBossRoomListIdx = this->final_boss_index;
+
```
ConnorForan, line 551: The DungeonGenerator should support [off-grid rooms](https://wofsauge.github.io/IsaacDocs/rep/enums/GridRooms.html).

Most likely by adding `DungeonGeneratorRoom offGridRooms[21];` (just flip the index)

It should maybe be possible for lua to place RoomConfig_Rooms into these slots (but only by placing RoomConfig_Rooms).

Now, if they were not populated by the mod, we must initialize the following [off-grid rooms](https://wofsauge.github.io/IsaacDocs/rep/enums/GridRooms.html) for (mostly) ALL dungeons:
- Error room (-2)
- Crawlspace dungeon (-4)
- Black market (-6)
- Members card secret shop (-13) (except home & ascent)
- Stairway to heaven angel shop (-18) (except home)
- The Rep+ lil portal room (-20)
- The Beast's room in Home?? (-10)

For `generate_dungeon` specifically we need these as well, regardless of floor with some exceptions (yes, the game always populates boss rush / mega satan / etc):
- Boss Rush (-5)
- Mega Satan (-7)
- Blue Womb trapdoor (-8) (for all stages except blue womb)
- Void trapdoor (-9) (for blue womb only)
- The "secret exit" room SPECIFICALLY for returning to the main path from Repentance alt path floors (-10)

These will (mostly) boil down to GetRandomRoom calls, but please refer to the 1.7.9b logic (easier to read, tho doesnt help for lil portal) to make sure we do this correctly. Ask for help in the discord thread if needed.

repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -5648,32 +5649,88 @@ HOOK_STATIC(LuaEngine, PostEntityKill, (Entity* ent) -> void, __stdcall) {
...
-			.push(DropPos, lua::Metatables::VECTOR)
-			.push(this, lua::Metatables::ENTITY_PLAYER)
-			.push(isGoldenTrinket)
-			.push(ReplaceTick)
-			.call(1);
+HOOK_METHOD(Level, generate_dungeon, (RNG* rng) -> void)
+{
+	bool skip = ProcessGenerateDungeonCallback(this, *rng, DEFAULT);
+	if (skip) {
+		return;
```
ConnorForan: I think we need to call generate_mirror_world here, for the appropriate stages. In theory it should just do what it normally does and flips the main layout... Guess we'll find out if it works.

ConnorForan: We'll need to confirm if there's anything else we need to do here since we're skipping the vanilla code, like anything the game changes or sets that isnt directly tied to the LevelGenerator portion.

I'll need to defer to Guantol for that, though.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
thicco-catto: Added support for all of these except for the last one "The "secret exit" room SPECIFICALLY for returning to the main path from Repentance alt path floors (-10)", I couldn't really see when this room spawns exactly (or at all). Could you specify further?

repentogon/LuaInterfaces/CustomCallbacks.cpp
thicco-catto: Turns out it actually works, rare nicalis W

repentogon/resources/scripts/main_ex.lua
thicco-catto: Yea, agree. Ended up implementing it exactly like you mentioned

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,84 @@
...
+	DungeonGenerationType generation_type;
+
+	bool reset_lil_portal = false;
+	int final_boss_index = -1;
+
+	DungeonGenerator(RNG* rng, Level* level, DungeonGenerationType generation_type, bool reset_lil_portal);
+
+	bool CanRoomBePlaced(XY& base_coords, int shape, int allowed_doors, bool allow_unconnected);
+
+	void BlockPositionsFromAllowedDoords(XY& base_coords, int shape, int allowed_doors);
```
Guantol-Lemat, line 53: Typo.
Also this is unused.

repentogon/Utils/LevelGenUtils.cpp
```diff
@@ -0,0 +1,294 @@
+#include "LevelGenUtils.h"
+#include <vector>
+#include <IsaacRepentance.h>
+
+#pragma region Helpers
+
+void PushCoordsIfValid(std::vector<XY>& list, XY& coords) {
```
Guantol-Lemat, line 7: This method should be static, as it is not exposed in the LevelGenUtils API

repentogon/Utils/LevelGenUtils.h
Guantol-Lemat, line 1: All methods here should be in their a "LevelGenUtils" namespace.

repentogon/Utils/LevelGenUtils.cpp
```diff
@@ -0,0 +1,294 @@
...
+	if (coords.ToGridIdx() >= 0) {
+		list.push_back(coords);
+	}
+}
+
+std::vector<XY> GetOccupiedCoords(XY& base_coords, int shape) {
+	std::vector<XY> occupied_coords = {};
```
Guantol-Lemat, line 14: Minor optimization: since we know the max amount of occupied coords, we can safely reserve enough space for the vector and avoid reallocations.

repentogon/Utils/LevelGenUtils.cpp
```diff
@@ -0,0 +1,294 @@
...
+		break;
+	default:
+		break;
+	}
+
+	return occupied_coords;
+}
+
+std::vector<XY> GetForbiddenNeighbors(XY& base_coords, int shape, int doors) {
+	std::vector<XY> forbidden_neighbors = {};
```
Guantol-Lemat, line 67: You can preallocate enough memory for this vector as well.

repentogon/Utils/LevelGenUtils.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+	switch (shape)
+	{
+	case ROOMSHAPE_1x1:
+		if ((doors & (1 << DOOR_SLOT_LEFT0)) == 0) {
```
Guantol-Lemat, line 72: Perhaps a utility `PushIfDoorNotAllowed` might help, given the many repeats.

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,84 @@
...
+	DungeonGeneratorRoom off_grid_rooms[20];
+	RNG* rng;
+	Level* level;
+	LevelGenerator level_generator;
+	DungeonGenerationType generation_type;
+
+	bool reset_lil_portal = false;
+	int final_boss_index = -1;
+
+	DungeonGenerator(RNG* rng, Level* level, DungeonGenerationType generation_type, bool reset_lil_portal);
```
Guantol-Lemat, line 49: Given the usage I think these methods can be set to private:
- CanRoomBePlaced
- PlaceRoom
- InitializeDefaultGridRooms
- CleanFloor
- PlaceRoomsInFloor
- ResetLevelGenerator

repentogon/LuaInterfaces/LuaDungeonGenerator.h
Guantol-Lemat, line 1: Both classes lack encapsulation.
This can be fine depending on the expected usage, tho it should be clear which functions/variables are meant to be implementation details, compared to something that's actually mutable/callable by outside systems.

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,84 @@
...
+enum DungeonGenerationType {
+	DEFAULT,
+	BLUE_WOMB,
+	BACKWARDS,
+	HOME,
+	RED_REDEMPTION,
+	GREED,
+};
+
+struct DungeonGeneratorRoom {
```
Guantol-Lemat, line 12: This whole struct can be made private, with DungeonGenerator as the only friend.

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,84 @@
...
+	int type;
+	int shape;
+	int minVariant;
+	int maxVariant;
+	int minDifficulty;
+	int maxDifficulty;
+	int subtype;
+	int mode;
+
+	int list_index;
```
Guantol-Lemat, line 28: Given the usage (only SetFinalBossRoom), I wonder if we could get rid of this, as it's a potential vector for bugs if later down the line something doesn't properly set this to a correct value.

In fact this value is invalid for offgrid rooms.

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,84 @@
...
+	DungeonGeneratorRoom();
+
+	DungeonGeneratorRoom(int list_index, uint32_t row, uint32_t col, int doors, RoomConfig_Room* room);
+	DungeonGeneratorRoom(int list_index, uint32_t row, uint32_t col, int doors, int stage, int type, int shape, int minVariant, int maxVariant, int minDifficulty, int maxDifficulty, int subtype, int mode);
+
+	RoomConfig_Room* GetRoomConfig(uint32_t seed, uint32_t required_doors);
+};
+
+struct DungeonGenerator {
+	DungeonGeneratorRoom rooms[169];
```
Guantol-Lemat, line 39: It should be documented that the rooms must always be synced with those in `level_generator`

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	return true;
+}
+
+bool DungeonGenerator::Generate() {
+	if (!this->ValidateFloor()) {
+		KAGE::_LogMessage(1, "[WARN] Failed to validate custom floor, not placing rooms.\n");
```
Guantol-Lemat, line 557: I would suggest having a unique header (like [LuaDungeonGenerator] or something similar) to make it clearer where the log comes from.

Also with `KAGE::_LogMessage`, there's no need to specify [WARN], as the type variable already prepends the message with it.

repentogon/resources/scripts/main_ex.lua
```diff
@@ -1855,12 +1855,27 @@ end
...
+-- I don't think we need these exposed anymore, but safer to just leave them alone since they were already exposed.
 rawset(Isaac, "RunPreRenderCallback", _RunPreRenderCallback)
 rawset(Isaac, "RunAdditiveCallback", _RunAdditiveCallback)
 rawset(Isaac, "RunEntityTakeDmgCallback", _RunEntityTakeDmgCallback)
 rawset(Isaac, "RunTriggerPlayerDeathCallback", _RunTriggerPlayerDeathCallback)
 
 
+local function RunPreGenerateDungeonCallback(callbackID, param, dungeonGenerator, rng, dungeonType)
+	for callback in GetCallbackIterator(callbackID, param) do
+		RunCallbackInternal(callbackID, callback, dungeonGenerator, rng, dungeonType)
```
Guantol-Lemat, line 1868: I'm thinking, should we store copy of the initial RNG values and reset the used RNG before running the next callback?

That way way failed generations do not alter the final outcome of the correct generation, or the vanilla generation if they all fail.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+		return false;
+	}
+
+	CleanFloor();
+
+	bool could_place_rooms = PlaceRoomsInFloor();
+	if (!could_place_rooms) {
+		KAGE::_LogMessage(1, "[WARN] Couldn't place the rooms in the level, clearing placed rooms...\n");
+		Reset();
```
Guantol-Lemat, line 568: Nitpick: It might be better to do an early return here, and avoid nesting the rest of the code in an `else` block.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	}
+	else {
+		if (!g_Game->IsGreedMode() && this->level->IsAltPath()) {
+			if (((this->level->_stage == STAGE1_1 && (this->level->GetCurses() & 2) != 0) || this->level->_stage == STAGE1_2)) {
+				this->level->generate_mirror_world();
+			}
+			else if (((this->level->_stage == STAGE2_1 && (this->level->GetCurses() & 2) != 0) || this->level->_stage == STAGE2_2)) {
```
Guantol-Lemat, line 575: iirc Mines is only generated when the player has the knife piece from MirrorWorld.

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,84 @@
...
+
+	void SetFinalBossRoom(DungeonGeneratorRoom* boss_room);
+
+	bool ValidateFloor();
+
+	void CleanFloor();
+
+	bool PlaceRoomsInFloor();
+
+	bool Generate();
```
Guantol-Lemat, line 79: Might be worth documenting that failed generation clears Level rooms.

repentogon/LuaInterfaces/CustomCallbacks.cpp
```diff
@@ -6402,4 +6403,93 @@ void CustomCallbacks::detail::ApplyPatches()
...
+		return false;
+	}
+
+	bool correctGeneration = generator.Generate();
+
+	return correctGeneration;
+}
+
+HOOK_METHOD(Level, generate_dungeon, (RNG* rng) -> void) {
+	bool skip = ProcessGenerateDungeonCallback(this, *rng, DEFAULT);
```
Guantol-Lemat, line 6444: Nitpick: Default should still use generationRNG, as the passed rng is meant to be a temporary variable used to get the initial room count; aside from that generationRNG is still used as the main RNG.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	this->shape = shape;
+	this->minVariant = minVariant;
+	this->maxVariant = maxVariant;
+	this->minDifficulty = minDifficulty;
+	this->maxDifficulty = maxDifficulty;
+	this->subtype = subtype;
+	this->mode = mode;
+}
+
+RoomConfig_Room* DungeonGeneratorRoom::GetRoomConfig(uint32_t seed, uint32_t required_doors) {
```
Guantol-Lemat, line 71: BUG: Two subsequent calls to this method would lead to different results, if the room uses a room filter.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	this->generation_type = generation_type;
+	this->reset_lil_portal = reset_lil_portal;
+
+	this->level_generator._rng = *rng;
+	this->level_generator._isChapter6 = false;
+	this->level_generator._isStageVoid = false;
+	this->level_generator._isXL = false;
+	
+	this->ResetLevelGenerator();
+	this->InitializeDefaultGridRooms();
```
Guantol-Lemat, line 112: `InitializeDefaultGridRooms` should be called at the end of generation, that way we are sure the rooms are initialized.

Currently there's nothing that can uninitialize these rooms, but future changes might violate this hidden invariant, and leave the stage in an "invalid" state.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+	if (this->CanRoomBePlaced(base_coords, room_config->Shape, doors, true)) {
+		return this->PlaceRoom(base_coords, doors, room_config);
+	}
+
+	return nullptr;
+}
+
+DungeonGeneratorRoom* DungeonGenerator::TryPlaceRoom(XY& base_coords, int doors, int stage, int type, int shape, int minVariant, int maxVariant, int minDifficulty, int maxDifficulty, int subtype, int mode) {
+	if (this->CanRoomBePlaced(base_coords, shape, doors, true)) {
```
Guantol-Lemat, line 201: Given that this is mostly a utility to both get a random room and place said room, I think we should immediately get the `RoomConfig_Room` and place it into the `DungeonGenerator`.

I think these would be beneficial for the following reasons:
- It removes the duplicate logic across the 3 Place methods.
- It simplifies the `DungeonGeneratorRoom` (a valid room is one that has roomConfig != nullptr;)
- It allows modders to handle the case in which no room matches a given filter (which is something that vanilla generation does), rather than delay the failure.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	int base_grid_index = base_coords.ToGridIdx();
+	if (!this->level_generator.IsPositionInBounds(base_coords)) {
+		return false;
+	}
+
+	if (!this->level_generator.is_pos_free(&base_coords, shape)) {
+		return false;
+	}
+
+	if (!allow_unconnected) {
```
Guantol-Lemat, line 125: `is_placement_valid` only checks if every neighbor has a valid connection (there is no neighbor that cannot connect to this position), so skipping this check only creates a possibly invalid layout, instead of allowing unconnected rooms.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+DungeonGeneratorRoom* DungeonGenerator::PlaceRoom(XY& base_coords, int doors, RoomConfig_Room* room_config) {
+	int col = base_coords.x;
+	int row = base_coords.y;
+
+	LevelGenerator_Room level_generator_room;
+	level_generator_room._gridColIdx = col;
+	level_generator_room._gridLineIdx = row;
+	level_generator_room._shape = room_config->Shape;
+	bool result = this->level_generator.place_room(&level_generator_room);
```
Guantol-Lemat, line 157: The result should either be checked, or you should follow this up with an assert to "document" that this is always expected to succeed.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+	LevelGenerator_Room placed_room = this->level_generator._rooms.at(this->level_generator._rooms.size() - 1);
```
Guantol-Lemat, line 159: You should take this room by reference.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	int col = base_coords.x;
+	int row = base_coords.y;
+	
+	LevelGenerator_Room level_generator_room;
+	level_generator_room._gridColIdx = col;
+	level_generator_room._gridLineIdx = row;
+	level_generator_room._shape = shape;
+	bool result = this->level_generator.place_room(&level_generator_room);
+
+	LevelGenerator_Room placed_room = this->level_generator._rooms.at(this->level_generator._rooms.size() - 1);
```
Guantol-Lemat, line 179: You should take this room by reference.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+DungeonGeneratorRoom* DungeonGenerator::PlaceOffGridRoom(int off_grid_index, int stage, int type, int shape, int minVariant, int maxVariant, int minDifficulty, int maxDifficulty, int subtype, int mode) {
+	int index = -off_grid_index - 1;
+
+	this->off_grid_rooms[index] = DungeonGeneratorRoom(-1, -1, -1, 0, stage, type, shape, minVariant, maxVariant, minDifficulty, maxDifficulty, subtype, mode);
+	DungeonGeneratorRoom* generatorRoom = &this->off_grid_rooms[index];
+
+	return generatorRoom;
+}
+
+DungeonGeneratorRoom* DungeonGenerator::TryPlaceDefaultStartingRoom(int doors) {
```
Guantol-Lemat, line 238: This doesn't account for Stage 6's unique starting room.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+	this->final_boss_index = -1;
+	for (size_t i = 0; i < 169; i++)
+	{
+		this->rooms[i] = DungeonGeneratorRoom();
+	}
+}
+
+bool DungeonGenerator::PlaceRoomsInFloor() {
+	this->level_generator.calc_required_doors();
```
Guantol-Lemat, line 514: Nitpick: Could be useful to comment that this is necessary to initialize the "required doors" in each `LevelGeneratorRoom`, which are then used to determine the allowed doors in the final `RoomDescriptor`.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+		);
+	}
+}
+
+void DungeonGenerator::SetFinalBossRoom(DungeonGeneratorRoom* boss_room) {
+	this->final_boss_index = boss_room->list_index;
+}
+
+bool DungeonGenerator::ValidateFloor() {
+	this->level_generator.calc_required_doors();
```
Guantol-Lemat, line 458: This is unnecessary here, unless I'm mistaken

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+	bool has_final_room = this->final_boss_index >= 0;
```
Guantol-Lemat, line 460: It might be better to just return early, since we treat a level that does not have a final room as invalid, which will in turn clear all placed rooms.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+	// Check if all rooms can fetch a room config
+	int initial_seed = this->rng->_seed;
+
+	for (LevelGenerator_Room room : this->level_generator._rooms)
+	{
+		DungeonGeneratorRoom generator_room = this->rooms[room._generationIndex];
+		RoomConfig_Room* room_config = generator_room.GetRoomConfig(this->rng->Next(), room._doors);
```
Guantol-Lemat, line 468: When a room is defined using a filter, `GetRoomConfig` has side effects that are not properly undone. If we remove the filter there is no issue with the current implementation, and we would also remove the need to undo side effects on the RNG.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	this->final_boss_index = boss_room->list_index;
+}
+
+bool DungeonGenerator::ValidateFloor() {
+	this->level_generator.calc_required_doors();
+
+	bool has_final_room = this->final_boss_index >= 0;
+
+	// Check if all rooms can fetch a room config
+	int initial_seed = this->rng->_seed;
```
Guantol-Lemat, line 463: It might be better to just create a copy of the RNG, that way it's guaranteed there are no side effects.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+		return luaL_error(L, "Invalid shape %d\n", shape);
+	}
+
+	int minVariant = (int)luaL_optinteger(L, 7, 0);
+	if (minVariant < 0) {
+		minVariant = 0;
+	}
+
+	int maxVariant = (int)luaL_optinteger(L, 8, -1);
+	if (maxVariant < minVariant && maxVariant >= 0) {
```
Guantol-Lemat, line 644: Not really a problem with this implementation, since this is a copy of the `GetRandomRoom` checks, but using `int` instead of `unsigned int` complicates the logic a bit. -1 is used as a way to quickly reach the maximum unsigned integer, making it "uncapped", rather than being a special value.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+	this->level_generator._rooms.clear();
+}
+
+void DungeonGenerator::Reset() {
+	this->CleanFloor();
+
+	this->ResetLevelGenerator();
+
+	this->final_boss_index = -1;
+	for (size_t i = 0; i < 169; i++)
```
Guantol-Lemat, line 507: Nitpick: Given that rooms need to be synchronized with the LevelGenerator rooms I would place this reset in `ResetLevelGenerator`.

repentogon/LuaInterfaces/LuaDungeonGenerator.cpp
```diff
@@ -0,0 +1,889 @@
...
+}
+
+void DungeonGenerator::ResetLevelGenerator() {
+	std::fill_n(this->level_generator._roomMap, 169, -1);
+	std::fill_n(this->level_generator._blockedPositions, 169, false);
+
+	this->level_generator._rooms.clear();
+}
+
+void DungeonGenerator::Reset() {
```
Guantol-Lemat, line 501: Offgrid rooms are not reset.

repentogon/LuaInterfaces/LuaDungeonGenerator.h
```diff
@@ -0,0 +1,84 @@
...
+	DungeonGeneratorRoom(int list_index, uint32_t row, uint32_t col, int doors, int stage, int type, int shape, int minVariant, int maxVariant, int minDifficulty, int maxDifficulty, int subtype, int mode);
+
+	RoomConfig_Room* GetRoomConfig(uint32_t seed, uint32_t required_doors);
+};
+
+struct DungeonGenerator {
+	DungeonGeneratorRoom rooms[169];
+	DungeonGeneratorRoom off_grid_rooms[20];
+	RNG* rng;
+	Level* level;
```
Guantol-Lemat, line 42: Is this necessary?

PR #768, opened 2025-12-02
Added SaveMenu.SetSaveSlotSpritesheet

thicco-catto:
Added new MC_POST_REPLACE_SAVE_SLOT_GRAPHICS callback. Passes the sprite of the save slot and the slot number thats being loaded. Can be used to make custom save slot sprites.

```lua
MyMod:AddCallback(ModCallbacks.MC_POST_REPLACE_SAVE_SLOT_GRAPHICS, function(_, sprite, slot)
    local persistentData = Isaac.GetPersistentGameData()

    local killedDeli = persistentData:IsBossKilled(BossType.DELIRIUM)

    if killedDeli then
        sprite:ReplaceSpritesheet(0, "gfx/ui/main menu/save drawings/kill_delirium.png")
        sprite:ReplaceSpritesheet(1, "gfx/ui/main menu/save drawings/kill_delirium.png", true)
    end
end)
```

### DIFF
--- a/repentogon/LuaInterfaces/Menus/LuaSaveMenu.cpp
+++ b/repentogon/LuaInterfaces/Menus/LuaSaveMenu.cpp
@@ -113,11 +113,23 @@ LUA_FUNCTION(Lua_SaveMenu_SetSlotSpritesheet)
 }
 
 void TryReplaceSlotGraphic(ANM2* sprite, std::string &customSprite) {
-	if (!customSprite.empty()) {
-		sprite->ReplaceSpritesheet(0, customSprite);
-		sprite->ReplaceSpritesheet(1, customSprite);
-		sprite->LoadGraphics(false);
+	if (customSprite.empty()) {
+		return;
 	}
+
+	bool success = sprite->ReplaceSpritesheet(0, customSprite);
+	if (!success) {
+		return;
+	}
+	
+	for (int i = 1; i < sprite->GetLayerCount(); i++) {
+		LayerState* layer = sprite->GetLayer(i);
+		if (layer != nullptr) {
+			layer->_visible = false;
+		}
+	}
+
+	sprite->LoadGraphics(false);
 }
 
 HOOK_METHOD(Menu_Save, replace_slot_graphics, (ANM2* sprite) -> void) {


jsgnextortex:
I honestly dont think this feature deserved a callback...wouldnt exposing the anm2s be enough?, the name of the callback alone is a dead giveaway that this is not supposed to be a callback...like a callback that only exists for a single function, shouldnt be a callback.

repentogon/LuaInterfaces/Menus/LuaSaveMenu.cpp
```diff
@@ -89,6 +89,49 @@ LUA_FUNCTION(Lua_SaveMenu_IsDeleteActive)
...
+	} else if (g_Manager->_currentSaveSlot == 2) {
+		CustomSaveSlotSprite2 = customSprite;
+	} else if (g_Manager->_currentSaveSlot == 3) {
+		CustomSaveSlotSprite3 = customSprite;
+	}
+
+	return 0;
+}
+
+void TryReplaceSlotGraphic(ANM2* sprite, std::string &customSprite) {
```
ConnorForan, line 115: nit: customSprite can be a const reference (`const std::string& customSprite`)

repentogon/LuaInterfaces/Menus/LuaSaveMenu.cpp
```diff
@@ -89,6 +89,49 @@ LUA_FUNCTION(Lua_SaveMenu_IsDeleteActive)
...
+	if (!customSprite.empty()) {
+		sprite->ReplaceSpritesheet(0, customSprite);
+		sprite->ReplaceSpritesheet(1, customSprite);
```
ConnorForan: If the game actually renders both layers at the same time for this custom sprite, they would overlap weirdly. The save drawings do seem to use transparency.

From taking a look at replace_slot_graphics, I think we should make sure layer 0 is visible, hide layer 1 (used for polaroid/negative), as well as hide all the mom's heart checkmark layers (2~12), just in case. So something like this:

You can get the LayerState via `sprite->GetLayer(i)`, though should sanity check that it did not return a nullptr before trying to use it. Good to be safe. I don't feel like we need to support the second layer atm.

Also, I think that ReplaceSpritesheet returns false if it fails? If that happens we should probably abort and not change anything further?

repentogon/LuaInterfaces/Menus/LuaSaveMenu.cpp
```diff
@@ -101,6 +144,7 @@ static void RegisterSaveMenuGame(lua_State* L)
 	lua::TableAssoc(L, "GetSelectedElement", Lua_SaveMenu_GetSelectedSave);
 	lua::TableAssoc(L, "SetSelectedElement", Lua_SaveMenu_SetSelectedSave);
 	lua::TableAssoc(L, "IsDeleteActive", Lua_SaveMenu_IsDeleteActive);
+	lua::TableAssoc(L, "SetSlotSprite", Lua_SaveMenu_SetSlotSprite);
```
ConnorForan: Maybe rename to SetSlotSpritesheet?

ConnorForan (commented):
Can the PR be renamed?

repentogon/LuaInterfaces/Menus/LuaSaveMenu.cpp
thicco-catto: Since ReplaceSpritesheet takes a non const reference I'd have to make a copy beforehand. Is this preferable?

ConnorForan: Ah, nevermind then, you can leave it

PR #797, opened 2025-12-29
Add Entity Save State Manager

Guantol-Lemat:
**INTERNAL**
- Adds the `EntitySaveStateManagement` system, which provides a way to extend the various entity save state structures (`GameStatePlayer`, `FamiliarData` and `EntitySaveState`) by assigning a unique ID to each active structure. These IDs can then be used by other systems to associate data with the base state.
- Adds the `EntityLifecycle` system, which notifies when an entity is "created" or "deleted".
    - Currently only used by the Lua `EntitySaveStateManager` to track of valid entities, delete entity data and quickly distinguish between valid and invalid userdata.

**LUA**
- Adds the `EntitySaveStateManager` global table to support persistent, modded save data for entities.
    - The data can be accessed through any `ModReference` obtained via `RegisterMod`, making the data "private" to each mod, unless explicitly shared by exposing the `ModReference` to the global scope.
    - Serialization and Deserialization are handled automatically by the system.

### DIFF
--- a/libzhl/functions/EntitySaveState.zhl
+++ b/libzhl/functions/EntitySaveState.zhl
@@ -5,29 +5,29 @@ __thiscall void EntitySaveState::constructor();
 __thiscall void EntitySaveState::destructor();
 
 struct EntitySaveState depends (Vector, KAGE_SmartPointer_EntitySaveState) {
-    int type : 0x0; // 0x0
-    int variant : 0x4; // 0x4
-    int subtype : 0x8; // 0x8
-	unsigned int intStorage1 : 0xc; // 0x1c
-	unsigned int intStorage2 : 0x10; // 0x10
-	unsigned int intStorage3 : 0x14; // 0x14
-    int intStorage4 : 0x18; // 0x18
-	int intStorage5 : 0x1c; // 0x1c
-    int16_t gridSpawnIdx : 0x20; // 0x20 // HIJACKED by ESSM
-	bool boolStorage1 : 0x22; // 0x22
-	bool boolStorage2 : 0x23; // 0x23
-	unsigned int intStorage6 : 0x24; // 0x24
-    Vector targetPosition : 0x28; // 0x28
-	unsigned int _intStorage7 : 0x30; // 0x30 // HIJACKED by ESSM
-    unsigned int _initSeed : 0x34; // 0x34
-	unsigned int _dropSeed : 0x38; // 0x34
-    int spawnerType : 0x3c; // 0x3C
-    int spawnerVariant : 0x40; // 0x40
-	float floatStorage1 : 0x50; // 0x50
-	float floatStorage2 : 0x54; // 0x54
-	unsigned int intStorage8: 0x5c; // 0x5c
-	int8_t byteStorage : 0x60; // 0x60
-	KAGE_SmartPointer_EntitySaveState entitySaveState : 0x6c; // 0x6c
+    int type : 0x0;
+    int variant : 0x4;
+    int subtype : 0x8;
+	unsigned int intStorage1 : 0xc;
+	unsigned int intStorage2 : 0x10;
+	unsigned int intStorage3 : 0x14;
+    int intStorage4 : 0x18;
+	int intStorage5 : 0x1c;
+    int16_t gridSpawnIdx : 0x20; // HIJACKED by ESSM
+	bool boolStorage1 : 0x22;
+	bool boolStorage2 : 0x23;
+	unsigned int intStorage6 : 0x24;
+    Vector targetPosition : 0x28;
+	unsigned int _intStorage7 : 0x30; // HIJACKED by ESSM
+    unsigned int _initSeed : 0x34;
+	unsigned int _dropSeed : 0x38;
+    int spawnerType : 0x3c;
+    int spawnerVariant : 0x40;
+	float floatStorage1 : 0x50;
+	float floatStorage2 : 0x54;
+	unsigned int intStorage8: 0x5c;
+	int8_t byteStorage : 0x60;
+	KAGE_SmartPointer_EntitySaveState entitySaveState : 0x6c;
 
 	{{
 		EntitySaveState() {


repentogon/LuaInit.cpp
```diff
@@ -172,16 +174,42 @@ void NukeConstMetatables(lua_State* L) {
...
 
+static void bind_lua_internals(lua_State* L, int tblIdx)
+{
+	int stackTop = lua_gettop(L);
+	tblIdx = lua_absindex(L, tblIdx);
+	if (!lua_istable(L, tblIdx))
+	{
+		ZHL::Log("[ERROR] _LuaBindings is not a table.\n");
+		assert(false);
+		return;
```
Sylmir, line 187: If this is fatal, just crash. Assertions are removed in debug, I'm not sure we want to continue in such a scenario.
You can invoke std;;terminate()

EDIT : considering we are in a Lua context, luaL_error is available, so maybe you can simply use that instead of terminating everything. Dunno how much sense that would make

EDIT 2 : Okay, luaL_error is not available here, diregard first edit

repentogon/LuaInit.cpp
```diff
@@ -172,16 +174,42 @@ void NukeConstMetatables(lua_State* L) {
...
+	}
+
```
Sylmir, line 189: LuaStackProtector in restricted scopes (using curly braces) can be used to achieve the same effect as the assertions

repentogon/LuaInit.cpp
```diff
@@ -172,16 +174,42 @@ void NukeConstMetatables(lua_State* L) {
...
+	lua_getfield(L, tblIdx, "ESSM");
+	EntitySaveStateManagement::detail::Init::BindLuaCallbacks(L, -1);
+	lua_pop(L, 1);
+	assert(lua_gettop(L) == stackTop);
+	lua_settop(L, stackTop);
+}
+
 HOOK_METHOD(LuaEngine, Init, (bool Debug) -> void) {
 	constexpr const char* C_BINDINGS_NAME = "_CBindings";
+	constexpr const char* LUA_BINDINGS_NAME = "_LuaBindings";
```
Sylmir: constexpr serves no purpose here

repentogon/LuaInit.cpp
```diff
@@ -199,6 +227,17 @@ HOOK_METHOD(LuaEngine, Init, (bool Debug) -> void) {
 	lua_getglobal(state, "_RunCallback");
 	g_LuaEngine->runCallbackRegistry->key = luaL_ref(state, LUA_REGISTRYINDEX);
 
+	stackTop = lua_gettop(L);
```
Sylmir: You can use the LuaStackProtector in a restricted scope to achieve the same result

repentogon/LuaInterfaces/LuaEntitySaveState.h
```diff
@@ -1,10 +1,18 @@
...
+#include "LuaCore.h"
 
 struct Lua_EntitySaveState {
     std::vector<EntitySaveState>* vec;
     int index;
+
+    static inline EntitySaveState& GetEntitySaveState(lua_State* L, int idx)
+    {
+        Lua_EntitySaveState* ud = lua::GetRawUserdata<Lua_EntitySaveState*>(L, idx, lua::metatables::EntitySaveStateMT);
+        EntitySaveState& st = (ud->vec->data()[ud->index]);
```
Sylmir: There is no need to use data() here, (*ud->vec)[ud->index] will work just a fine. If this is an operation you do multiple times, having a getter in Lua_EntitySaveState is a good alternative. You can also remove the parentheses

Also, considering we are in a Lua binder, you can add a bound check if you want, luaL_error is available

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+    static void RestoreEntity(Entity& entity, uint32_t id)
+    {
+        s_luaCallbacks.RestoreEntity(entity, id);
+    }
+}
+
+// Operations API
+namespace ESSM
+{
+    void EntitySaveState_ClearVector(std::vector<EntitySaveState>& vector)
```
Sylmir: Rename this function to ClearBatch to avoid the implication that the function will clear the vector passed as parameter. In addition, if the function does not modify the vector, make it a const reference

repentogon/SaveStateManagement/EntitySaveStateManagement.h
```diff
@@ -0,0 +1,109 @@
...
+
+    uint32_t& GameStatePlayer_GetImmaculateConceptionState(GameStatePlayer& data);
+    uint32_t& GameStatePlayer_GetCambionConceptionState(GameStatePlayer& data);
+
+    uint32_t& FamiliarData_GetState(FamiliarData& data);
+    uint32_t& FamiliarData_GetRoomClearCount(FamiliarData& data);
+
+    // Clears all ids from the passed vector
+    // DOES NOT CLEAR THE VECTOR, IT MUST BE MANUALLY CLEARED AFTERWARDS!!!
+    void EntitySaveState_ClearVector(std::vector<EntitySaveState>& vector);
```
Sylmir: See my comment in the definition in the .cpp file

repentogon/LuaInterfaces/LuaEntitySaveState.cpp
```diff
@@ -162,16 +165,16 @@ struct Lua_EntitySaveStateAPI : Lua_EntitySaveState {
 	LUA_FUNCTION(Lua_GetGridSpawnIdx) {
 		Lua_EntitySaveStateAPI* ud = GetData(L, 1);
 		EntitySaveState& st = (ud->vec->data()[ud->index]);
-		lua_pushinteger(L, st.gridSpawnIdx);
+		lua_pushinteger(L, ESSM::EntitySaveState_GetGridSpawnIdx(st));
```
Sylmir, line 168: General comment : I do not understand why there's a mix of C-style functions to work on fields of structures and C++-style methods. Either pick one, or explain why it is the way that it is (applies to basically every occurrence of EntitySaveState_GetSomething)

repentogon/LuaInterfaces/_Internals.cpp
```diff
@@ -16,7 +17,31 @@ LUA_FUNCTION(Lua_RaiseModError)
...
+	}
+
+	lua_pushstring(L, modRef->_mainLuaPath.c_str());
+	return 1;
+}
+
 void LuaInternals::RegisterInternals(lua_State *L)
 {
     lua::TableAssoc(L, "RaiseModError", Lua_RaiseModError);
+	lua::TableAssoc(L, "GetModId", Lua_GetModId);
```
Sylmir, line 42: Not sure about the name, it is rather uncommon for strings to be IDs

repentogon/LuaInterfaces/_Internals.cpp
```diff
@@ -16,7 +17,31 @@ LUA_FUNCTION(Lua_RaiseModError)
...
+
+	lua_newtable(L);
+	EntitySaveStateManagement::detail::Init::RegisterLuaInternals(L);
```
Sylmir: Move the lua_newtable(L) inside RegisterLuaInternals as it makes the function more self-contained (unless you explicitly want the function to be usable on a table already on the stack, in which case make it clear at the function declaration level)

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+void EntityLifecycle::detail::Init::BindLuaCallbacks(lua_State* L, int tblIdx)
+{
+    assert(L == g_LuaEngine->_state);
+
+    tblIdx = lua_absindex(L, tblIdx);
+    if (!lua_istable(L, tblIdx))
+    {
+        ZHL::Log("[ERROR] EntityLifecycle::BindLuaCallbacks: expected a table\n");
+        assert(false);
```
Sylmir, line 26: Same as above, std::terminate instead of assert (unless the program can continue execution in a broken state)

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+    Data::s_luaNewEntityCallback = LUA_NOREF;
+    Data::s_luaDeleteEntityCallback = LUA_NOREF;
+
+    auto bindCallback = [](lua_State* L, int tblIdx, const char* fieldName, int& outRef)
+    {
+        lua_getfield(L, tblIdx, fieldName);
+        if (!lua_isfunction(L, -1))
+        {
+            ZHL::Log("[ERROR] EntityLifecycle::BindLuaCallbacks: Expected '%s' to be a function\n", fieldName);
+            lua_pop(L, 1);
```
Sylmir, line 42: As usual, std::terminate may be more appropriate than assert

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+    bindCallback(L, tblIdx, "NewEntity", Data::s_luaNewEntityCallback);
+    bindCallback(L, tblIdx, "DeleteEntity", Data::s_luaDeleteEntityCallback);
+}
+
+namespace EntityLifecycle::Core
+{
+    static void NewEntity(Entity& entity)
+    {
+        LuaEngine* lua = g_LuaEngine;
+        if (!lua)
```
Sylmir: Can this really happen ?

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+        {
+            ZHL::Log("[ERROR] EntityLifecycle::NewEntity Lua error: %s\n", lua_tostring(L, -1));
+            lua_pop(L, 1);
+        }
+    }
+
+    static void DeleteEntity(Entity& entity)
+    {
+        LuaEngine* lua = g_LuaEngine;
+        if (!lua)
```
Sylmir: Again

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+    return entity;
+}
+
+// Quick way of deleting entities overridden by EntityFactory
+HOOK_METHOD(Entity, Remove, () -> void)
+{
+    super();
+    if (this->_removedByFactory)
+    {
```
Sylmir, line 103: Is it safe to fire the callback after the call to super() ? Are we sure nothing changes inside the entity when it is removed, all fields accessible on the Lua side will have the expected value ? Though I do understand that there is a need to check whether the factory effectively removed the entity before firing the callback, so a bit of a chicken-and-egg problem here

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+        outRef = luaL_ref(L, LUA_REGISTRYINDEX);
+    };
+
+    bindCallback(L, tblIdx, "NewEntity", Data::s_luaNewEntityCallback);
+    bindCallback(L, tblIdx, "DeleteEntity", Data::s_luaDeleteEntityCallback);
+}
+
+namespace EntityLifecycle::Core
+{
+    static void NewEntity(Entity& entity)
```
Sylmir: Retype the parameter to void* or uintptr_t to clearly show that you are not pushing a userdata reference, but simply an abstract pointer

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+        lua_pushinteger(L, lua->GetMaskedPointer((uintptr_t)&entity));
+
+        if (lua_pcall(L, 1, 0, 0) != LUA_OK)
+        {
+            ZHL::Log("[ERROR] EntityLifecycle::NewEntity Lua error: %s\n", lua_tostring(L, -1));
+            lua_pop(L, 1);
+        }
+    }
+
+    static void DeleteEntity(Entity& entity)
```
Sylmir: Retype the parameter just like with NewEntity

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+static void Patch_EntityListDestructor_RemoveMainEL()
+{
+    intptr_t addr = (intptr_t)sASMDefinitionHolder->GetDefinition(&AsmDefinitions::EntityList_destructor_RemoveEntityMainEL);
+    ZHL::Log("[REPENTOGON] Patching EntityList::destructor for EntityLifecycle at %p\n", addr);
+
+    ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::GP_REGISTERS_STACKLESS, true);
+    ASMPatch patch;
+
+    intptr_t resumeAddr = addr + 8;
+    constexpr size_t RESTORED_BYTES = 3;
```
Sylmir: constexpr serves no purpose here

repentogon/Patches/EntityLifecycle.cpp
ConnorForan: I don't quite understand this, the callsites for this function looks like it has an Entity* received from the game, not from lua?

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+static void Patch_EntityListDestructor_RemovePersistentEL()
+{
+    intptr_t addr = (intptr_t)sASMDefinitionHolder->GetDefinition(&AsmDefinitions::EntityList_destructor_RemoveEntityPersistentEL);
+    ZHL::Log("[REPENTOGON] Patching EntityList::destructor for EntityLifecycle at %p\n", addr);
+
+    ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::GP_REGISTERS_STACKLESS, true);
+    ASMPatch patch;
+
+    intptr_t resumeAddr = addr + 8;
+    constexpr size_t RESTORED_BYTES = 3;
```
Sylmir: constexpr serves no purpose here

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+static void Patch_EntityListReset_RemoveNonPersistentEntity()
+{
+    intptr_t addr = (intptr_t)sASMDefinitionHolder->GetDefinition(&AsmDefinitions::EntityList_Reset_RemoveNonPersistentEntity);
+    ZHL::Log("[REPENTOGON] Patching EntityList::Reset for EntityLifecycle at %p\n", addr);
+
+    ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::GP_REGISTERS_STACKLESS, true);
+    ASMPatch patch;
+
+    intptr_t resumeAddr = addr + 8;
+    constexpr size_t RESTORED_BYTES = 3;
```
Sylmir: constexpr

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+    patch.AddBytes(ByteBuffer().AddAny((void*)addr, RESTORED_BYTES)) // Moves the entity into ECX
+        .PreserveRegisters(savedRegisters)
+        .AddInternalCall(asm_remove_and_delete_entity)
+        .RestoreRegisters(savedRegisters)
+        .AddRelativeJump((void*)resumeAddr);
+
+    sASMPatcher.PatchAt((void*)addr, &patch);
+}
+
+static void Patch_EntityListReset_RemoveNonPersistentEntity()
```
Sylmir, line 157: Unless I'm mistaken, this is the third function in a row that does more or less the same thing, barring the address of the patch ID and the string identifier in the log, maybe make a single higher level function that is called three times with different parameters ?

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,255 @@
...
+{
+    intptr_t addr = (intptr_t)sASMDefinitionHolder->GetDefinition(&AsmDefinitions::EntityList_Update_RemoveEntityMainEL);
+    ZHL::Log("[REPENTOGON] Patching EntityList::Update for EntityLifecycle at %p\n", addr);
+
+    ASMPatch::SavedRegisters savedRegisters(ASMPatch::SavedRegisters::Registers::GP_REGISTERS_STACKLESS, true);
+    ASMPatch patch;
+
+    intptr_t resumeAddr = addr + 7;
+
+    patch.AddBytes("\x8B\xCA") // MOV ECX, EDX
```
Sylmir: Don't we have a function for register to register moves ? If not I should add one

repentogon/Patches/EntityLifecycle.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "LuaCore.h"
+
+namespace EntityLifecycle
```
Sylmir: You can use inline namespaces in C++14 and above. namespace EntityLifeCycle::detail::Init is valid

repentogon/Patches/GameStateManagement.cpp
```diff
@@ -201,12 +210,17 @@ HOOK_METHOD(GameState, write, (GameStateIO** io) -> bool)
...
 }
 
 HOOK_METHOD(GameState, write_rerun, (GameStateIO** io) -> bool)
 {
-	if (!super(io))
+	ESSM::detail::SaveData::WriteState state = ESSM::detail::SaveData::WriteGameState();
+	bool success = super(io);
+	ESSM::detail::SaveData::RestoreWrittenStates(state);
+
+	if (!success)
```
Sylmir, line 251: Unrelated, but GH won't let me comment on something you didn't change (thanks GH...)

Line 229 : std::aoi is unsafe as it does not handle out of range values, std::strtol is much safer

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+    struct Data
+    {
+        std::vector<uint32_t> reusableIds;
+        uint32_t totalIds = 0;
+        std::vector<HijackedState> hijackedStates; 
+    };
+
+    class LuaCallbacks
+    {
+        private: int m_storeEntity = LUA_NOREF;
```
Sylmir: You do not need to have private: and public: before every attribute or method. private, protected and public are scope based and persist until any of them is encountered

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+
+                func(player->_unlistedRestoreState, std::forward<Args>(args)...);
+            };
+
+            for (Entity_Player* player : players)
+            {
+                handlePlayer(player);
+            }
+
+            for (size_t i = 0; i < 4; i++)
```
Sylmir: Do not hardcode 4 ; if the accessed array is static, sizeof can give you the size at compile time

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+
+                detail::iterate_vector(player->_unlistedRestoreState._movingBoxContents, func, std::forward<Args>(args)...);
+            };
+        
+            for (Entity_Player* player : players)
+            {
+                handlePlayer(player);
+            }
+        
+            for (size_t i = 0; i < 4; i++)
```
Sylmir: Hardcoded 4

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+}
+
+constexpr size_t ESAU_JR_PLAYERS = 4;
+
+namespace ESSM::PlayerIterators
+{
+    namespace InGame
+    {
+        template <typename Func, typename... Args>
+        static void All(Func&& func, Args&&... args)
```
Sylmir: All in C++ is associated with std::all, which is a boolean function. If you want to apply a function on all elements of a structure, for_each / each is preferred (applies to basically all occurrences of All as part of a function name here)

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+            {
+                func(entity, std::forward<Args>(args)...);
+            }
+        }
+    }
+
+    namespace Structure
+    {
+        template <typename Func, typename... Args>
+        static void BackwardsStage(BackwardsStageDesc& backwardsStage, Func&& func, Args&&... args)
```
Sylmir: Please rename this function to something that does not read like a name of a type (EachBackwardStage ?)

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+        using Data = typename Traits::DataType;
+        using Marker = typename Traits::MarkerType;
+        using Id = typename Traits::IdType;
+
+        private: static constexpr Marker CLEARED_MARKER = 0x5249;
+        private: static constexpr Marker READ_MARKER = 0x5248;
+        private: static constexpr Marker HIJACK_MARKER = 0x5247;
+        private: static constexpr Marker WRITTEN_MARKER = 0x5246;
+
+        private: static const Marker& get_marker_value(const Data& data)
```
Sylmir: Please remove the non const version of get_marker_value and add `void set_marker_value(Data& data, Marker marker).

Right now the only reason you need to have two versions, with one that performs const_cast to remove the const added by accessing an as_const reference's subobject, is because you want to write get(X) = Y. set(X, Y) is perfectly fine and more readable (and more in-line with how we do pretty much everything else)

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+        {
+            return data.*(Traits::marker_member);
+        }
+
+        private: static Marker& get_marker_value(Data& data)
+        {
+            return const_cast<Marker&>(get_marker_value(std::as_const(data)));
+        }
+
+        private: static const Id& get_id_value(const Data& data)
```
Sylmir: Same remarks as with get_marker_value

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+namespace ESSM
+{
+    template <typename Traits>
+    class HijackManager
+    {
+        using Data = typename Traits::DataType;
+        using Marker = typename Traits::MarkerType;
+        using Id = typename Traits::IdType;
+
+        private: static constexpr Marker CLEARED_MARKER = 0x5249;
```
Sylmir: This implicitly establishes that MarkerType in any Traits shall be an (possibly signed) integral type with a width of at least two bytes. Document that restriction somewhere, as it can easily lead to errors if someone uses a smaller marker. You can also add a `static_assert(std::is_integral_v<Marker> && sizeof(Marker) >= 2);` to enforce this at compile-time

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+            return id;
+        }
+    };
+
+    struct EntityTraits {
+        using DataType = EntitySaveState;
+        using MarkerType = short;
+        using IdType = uint32_t;
+
+        static constexpr MarkerType DataType::* marker_member = &EntitySaveState::gridSpawnIdx;
```
Sylmir: This is incredibly cursed. Just use a function called GetMarkerMember() and a function called GetIdMember(). You can write a template interface that is implemented by a given trait to enforce this easily.

```cpp
template<typename MarkerR, typename IDR, typename TDataType>
struct ITraitsGetters {
    using DataType = TDataType;
    using Marker = MarkerR;
    using IdType = IDR;

    virtual MarkerR* GetMarker(DataType&) = 0;
    virtual IDR* GetID(DataType&) = 0;
};

struct EntityTraits : public ITraitsGetter<short, uint32_t, EntitySaveState> {
    short* GetMarker(EntitySaveState& s) { return &s.gridSpawnIdx; }
    uint32_t* GetID(EntitySaveState& s) { return &s._intStorage7; }
};
```

You're already using function pointers. The call overhead of a virtual method is the same, as you need to move the pointer in a register either way, and call on a register allows pointer displacement.

This works for all traits, just use the appropriate type parameters when configuring the parent class (basically moving them in the parent class).

In addition, constexpr serves no real purpose here.

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+    template<typename T>
+    struct TraitsFor;
+
+    struct EntityTraits
+    {
+        using Hijack = EntityHijackManager;
+        static constexpr bool can_be_cleared = false;
+        static constexpr bool can_recurse = true;
+
+        static EntitySaveState* recurse(EntitySaveState& s)
```
Sylmir: If I encounter a function called recurse, I'm expecting it to perform some kind of recursive computation. Maybe rename this to something more intuitive ? GetRecursiveChild ?

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+{
+    void LuaCallbacks::BindLuaCallbacks(lua_State* L, int tblIdx)
+    {
+        assert(L == g_LuaEngine->_state);
+    
+        tblIdx = lua_absindex(L, tblIdx);
+        if (!lua_istable(L, tblIdx))
+        {
+            ZHL::Log(LOG_ERROR_HEADER "ESSM::BindLuaCallbacks: expected a table\n");
+            assert(false);
```
Sylmir, line 751: std::terminate if necessary

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+            return;
+        }
+    
+        luaL_unref(L, LUA_REGISTRYINDEX, this->m_storeEntity);
```
Sylmir: I've seen enough sequences of luaL_unref into resetting the key to LUA_NOREF that I think it warrants a dedicated function for extra safety

repentogon/SaveStateManagement/EntitySaveStateManagement.h
```diff
@@ -0,0 +1,109 @@
...
+        void BindLuaCallbacks(lua_State* L, int tbl);
+    }
+
+    namespace SaveData
+    {
+        using StatePtr = std::variant<GameStatePlayer*, FamiliarData*, EntitySaveState*>;
+
+        struct _WriteState
+        {
+            std::vector<std::pair<StatePtr, uint32_t>> writtenEntitySaveStates;
```
Sylmir: You may want to replace the pairs with structures containing fields with explicit names (using std::pair outside of the context of a map is a bit weird)

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+            std::error_code errorCode;
+            if (!std::filesystem::exists(filePath, errorCode))
+            {
+                continue;
+            }
+
+            lua_rawgeti(L, LUA_REGISTRYINDEX, modRef._luaTableRef->_ref);
+            lua_rawseti(L, mods, i + 1);
+
+            std::string id = modRef._mainLuaPath;
```
Sylmir: Use a (const) reference to avoid a copy

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
```diff
@@ -0,0 +1,3160 @@
...
+                stringSizes[i] = std::strlen(strings[i]);
+                stringSize += NEW_LINE_SIZE + stringSizes[i] + 1; // newline character
+            }
+
+            stringSize += FOOTER_SIZE;
+ 
+            std::unique_ptr<char[]> consoleString = std::make_unique<char[]>(stringSize + 1); // + 1 for the null terminator
+            size_t consoleStringPos = 0;
+
+            std::memcpy(consoleString.get(), HEADER, HEADER_SIZE);
```
Sylmir: If you are working with strings, use strcat instead of memcpy, this will avoid the risks of overwriting data or moving too far when concatenating strings

repentogon/SaveStateManagement/EntitySaveStateManagement.h
```diff
@@ -0,0 +1,109 @@
...
+
+            private: _WriteState() = default;
+
+            friend struct WriteState;
+        };
+
+        struct _ReadState
+        {
+            std::vector<StatePtr> readEntitySaveStates;
+            std::vector<std::pair<uint32_t, uint32_t>> restoreEntityIdPairs; // readId, id (for Lua Callbacks)
```
Sylmir: Same as above for replacing pair with a struct

Sylmir (commented):
Gave up midway through EntitySaveStateManagement.cpp (it's too long), here's my feedback for what comes before and a bit of what comes after

repentogon/LuaInit.cpp
Guantol-Lemat: `assert(false)` was placed so that the game would crash specifically on debug builds, technically the game can still continue even if we cannot bind the lua callbacks.
That being said it is also true that this would only happen if something went wrong in main_ex, so maybe we should call terminate, given that most systems that rely on the bindings would not work.

repentogon/LuaInterfaces/LuaEntitySaveState.cpp
Guantol-Lemat: I could add them to a namespace based on the structure name, but that would mean having to specifiy that the structures come from the global scope in the functions.
I may be wrong, but they can't be added to the EntitySaveState structure directly since those belong to `libzhl` and I get either a link error if I try to place the definition in `IsaacRepentance_static.cpp` (since EntitySaveStateManagement.cpp doesn't get compiled) or a `warning C4273: inconsistent dll linkage` if I define it in `zhlREPENTOGON`, which I assume would cause problems.

repentogon/Patches/EntityLifecycle.cpp
Guantol-Lemat: All I see when searching for Move is:
- MoveFromMemory() equivalent of MOV src, [dst+offset]
- MoveToMemory() equivalent of MOV [dst+offset], src (I assume because the documentation says it allows moves in the same form as MoveFromMemory()
- MoveImmediate() equivalent of MOV dst, imm32

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
Guantol-Lemat: I know, this is deliberate. I prefer making the visibility explicit for each field, but I can still change it if needed.

repentogon/Patches/EntityLifecycle.cpp
Guantol-Lemat: Yes, remove simply "flags" the entity for removal (which is applied towards the end of EntityList::Update), then it runs the MC_POST_ENTITY_REMOVE callback.
If we were to invoke it before super() there would be problems caused by the entity being treated as invalid during the execution of the callback.

repentogon/LuaInterfaces/_Internals.cpp
ConnorForan: Strings are the closest thing we can have to an identifier for mods, though there are a few different things that could mean. Still, in this context, this is an internal-only function used to obtain the unique/consistent identifier for per-mod savedata.

It seems like we are returning `_mainLuaPath`. What does that look like in practice? I assume that the path is relative at least, but does it actually include `main.lua`? If RegisterMods is called multiple times by one mod, do they all have the same value for this?

I assume that the directory name by itself, OR the directory name of the `data` folder, could be sufficient, though again depends on the nuances of what gets populated as `_mainLuaPath`.

Though I do believe that `_mainLuaPath` works just fine as is for our current purposes, just making some considerations.

repentogon/resources/scripts/repentogon_extras/entity_save_state_manager.lua
```diff
@@ -0,0 +1,624 @@
...
+
+---Set of entities PtrHash that exist
+---@type table<integer, true|nil>
+local s_validEntities = {}
+
+---Persistent mod data for existing entities
+---@type table<string, table<integer, table>>
+local s_entityData = {}
+
+---Persistent mdo data for entity save states
```
ConnorForan: mdo

repentogon/Patches/EntityLifecycle.cpp
ConnorForan: I believe that CopyRegister is the "MOV" equivalent

repentogon/Patches/EntityLifecycle.cpp
```diff
@@ -0,0 +1,247 @@
...
+        outRef = luaL_ref(L, LUA_REGISTRYINDEX);
+    };
+
+    bindCallback(L, tblIdx, "NewEntity", Data::s_luaNewEntityCallback);
+    bindCallback(L, tblIdx, "DeleteEntity", Data::s_luaDeleteEntityCallback);
+}
+
+namespace EntityLifecycle::Core
+{
+    static void NotifyNewEntity(uintptr_t entityPtr)
```
ConnorForan: It's not a massive deal, and I know it was Sylmir's suggestion, but personally I don't know if i agree with the use of uintptr_t as the parameter here

It's true that its the minimal required input in order to push to lua, but since the only acceptable input is a valid Entity object, the original parameter type of Entity& made sense to me.

This may be one of those subjective things though. I understand the possible sentiment for reducing it to uintptr_t since that's all you really "need", but Entity& is a much more restrictive input that makes it most unlikely (albeit not impossible) to provide an invalid parameter, and also excludes nullptr equivalents. So I guess its specificity vs minimalism? Probably depends on which angle you look at it from, but from an external perspective Entity& made sense to me.

repentogon/SaveStateManagement/EntitySaveStateManagement.cpp
ConnorForan: I'm with sylmir on this one in terms of standards, private: and public: preceding blocks of fields/functions is clear enough imo, especially when oftentimes all fields are private and (most) functions are public, with private functions getting their own block if necessary. If private and public definitions were extremely fine-grained, there would probably be something wrong with the design, but that is not the case here, its relatively clean-cut.

repentogon/LuaInterfaces/_Internals.cpp
Guantol-Lemat: `_mainLuaPath` is used because that's what the game uses to identify mod references that belong to the same mod when rerunning a mod's script (through `luamod` or other) in order to delete them.

Granted, they most likely used it since it was easier to compare the value passed to `RunScript` with `_mainLuaPath` for a direct match than to use any of the others. However, given that this is `_mainLuaPath`'s only purpose, it felt like the closest thing to an ID for mod references.

Also, yes, `_mainLuaPath` includes the `main.lua` part.

PR #801, opened 2026-01-04
Add RoomConfig(Set)::LoadStb

ConnorForan:
Allows virtual rooms to be loaded from a content folder stb file. Otherwise behaves the same as existing virtual room support provided for the luaroom format.

The filepath provided to the function is relative, starting from `<mod>/content(-repentogon)/rooms/`. If multiple mods have matching files, they will all be loaded. This is consistent behavior for content files, and mods will be somewhat responsible for avoiding collisions.

The game's own logic for reading rooms from an stb file was able to be used for this purpose, though we need to make a copy of it so that we own the memory allocated for the Spawns and such.

Small note for a change introduced by this PR: I am pretty confident that `RoomSpawn::CountEntries` is `uint8_t`, not `uint16_t`. I had an issue with this due to using the game's own code to generate the spawn data. I've updated the zhl accordingly. For evidence, refer to the assembly for `RoomConfig::Spawn::Entries` (byte ptr). `uint16_t` is correct for `RoomConfigRoom::SpawnCount` though (word ptr).

### DIFF
--- a/repentogon/Patches/VirtualRoomSets.cpp
+++ b/repentogon/Patches/VirtualRoomSets.cpp
@@ -1248,15 +1248,11 @@ static std::vector<std::optional<RoomConfig_Room>> build_rooms(lua_State* L, int
 	return rooms;
 }
 
-static void build_lua_return_table(lua_State* L, int index, std::vector<RoomConfig_Room*>& placedRooms)
+static void build_lua_return_table(lua_State* L, std::vector<RoomConfig_Room*>& placedRooms)
 {
-	int absIndex = lua_absindex(L, index);
-	size_t size = (size_t)lua_rawlen(L, absIndex);
-	assert(placedRooms.size() == size); // placed rooms's size must be equal to size of the table
-
 	lua_newtable(L);
 
-	for (size_t i = 1; i <= size; i++)
+	for (size_t i = 1; i <= placedRooms.size(); i++)
 	{
 		auto* room = placedRooms[i - 1];
 		if (!room)
@@ -1283,7 +1279,7 @@ void VirtualRoomSetManager::__AddLuaRooms(lua_State* L, uint32_t stageId, int mo
 	}
 
 	LogUtility::LogContext logContext;
-	logContext.emplace_back(REPENTOGON::StringFormat("[LUA] Add Rooms to set (%u, %i): ", stageId, mode));
+	logContext.emplace_back(REPENTOGON::StringFormat("[LUA] Add Lua Rooms to set (%u, %i): ", stageId, mode));
 	std::string luaCaller = LogUtility::Lua::GetStackLevelInfo(L, 1);
 	if (!luaCaller.empty())
 	{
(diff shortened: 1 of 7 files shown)


repentogon/Patches/VirtualRoomSets.cpp
```diff
@@ -1299,7 +1295,100 @@ void VirtualRoomSetManager::__AddLuaRooms(lua_State* L, uint32_t stageId, int mo
...
+
+	for (ModEntry* mod : g_Manager->GetModManager()->_mods)
+	{
+		if (!mod->IsEnabled()) continue;
+
+		std::string contentPath;
+		mod->GetContentPath(&contentPath, &filename);
+
+		if (const char* expandedPath = g_ContentManager.GetMountedFilePath(contentPath.c_str())) {
+			modfilepaths.push_back(expandedPath);
```
Guantol-Lemat, line 1330: I wonder if we should add a helper function to MiscFunctions that collects all paths for a given content path, instead of handling this manually each time. Primarily due to the risk of a memory leak if someone forgets to delete/free the expandedPath.

repentogon/Patches/VirtualRoomSets.cpp
```diff
@@ -1299,7 +1295,100 @@ void VirtualRoomSetManager::__AddLuaRooms(lua_State* L, uint32_t stageId, int mo
 		LogUtility::PrintConsole(logType, luaCaller + ": Something went wrong whilst adding rooms to a room set. Check repentogon.log for more details.");
 	}
 
-	build_lua_return_table(L, absTableIndex, placedRooms);
+	assert((size_t)lua_rawlen(L, absTableIndex) == placedRooms.size()); // placed rooms's size must be equal to size of the table
+
+	build_lua_return_table(L, placedRooms);
+}
+
+void VirtualRoomSetManager::__AddStbRooms(lua_State* L, uint32_t stageId, int mode, std::string filename) {
```
Guantol-Lemat, line 1303: Not necessary currently (since this is marked as a "detail" function), but maybe this should be turned into a general AddStbRooms helper.

Right now the function is tied to Lua, but the only uses of `lua_State` seem to be adding context to the logs and returning the placed rooms as a table. both of which could be left to the caller. Unless we don’t expect to always load STBs this way, in which case keeping it Lua-specific makes sense.

repentogon/Patches/VirtualRoomSets.cpp
```diff
@@ -1299,7 +1295,100 @@ void VirtualRoomSetManager::__AddLuaRooms(lua_State* L, uint32_t stageId, int mo
...
+			bufferRoomSet._configs = new RoomConfig_Room[numRooms];
+			bufferRoomSet._count = numRooms;
+
+			rooms.reserve(rooms.size() + numRooms);
+
+			for (int i = 0; i < numRooms; i++)
+			{
+				RoomConfig::read_room(&bufferRoomSet, i, stageId, file);
+				// This will create a copy of the RoomConfig_Room, intentional since the copy's RoomSpawns will use memory allocated by us instead of the game.
+				rooms.push_back(bufferRoomSet._configs[i]);
```
Guantol-Lemat, line 1371: I'm fine with the current use of `read_room`, since it doesn't change global state. That said, we may want to revisit this in the future, as it currently requires copying the read rooms twice (once here and again when adding to the manager).

Still, this isn’t something I’d focus on for this PR, since there are plans to change how rooms are stored and added—potentially allowing direct writes into the buffer and avoiding copies altogether.

ConnorForan: The copies are honestly negligible in the grand scheme of things here, but I get you.

I tried to address it a bit in the comments, but I ran into a problem where if we attempt to move or destroy a RoomConfig_Room allocated by the game's own code, it corrupts the heap. It's related to memory allocation and compiler stuff, we can't simply call `delete` on a set of spawns allocated by the game, our zhl code for destruction only works on the things we own. The game does not have easily available functions for creating and destroying rooms, as you are probably aware of since you had to implement code for construction/destruction manually.

As for why a room needs to be deleted at all, it doesn't really, but moves and such can also end up freeing memory or calling destructors, so I still ran into heap corruptions here and there. The way I have it set up now works fine, at least, since we own our copies and let the game clean up its own memory.

If you rework things in the future one suggestion could be to switch to pointers or smart pointers for the RoomConfig_Rooms so that you can mostly avoid copies, moves, deletes etc. Though I guess smart pointers could still attempt to destroy stuff in some cases.

For the record, I don't personally think any of this is worth switching to a reimplementation of read_room. At least not until we have some feature justification for doing so.

repentogon/Patches/VirtualRoomSets.cpp
ConnorForan: Good callout. Personally I'd completely replace our version of GetMountedFilePath with one that returns an std::string (making the 'real' one private or something). And yes I agree, a helper to gather the content paths would be a good addition as well.

I can do these things, but probably separately from the PR.

repentogon/Patches/VirtualRoomSets.cpp
ConnorForan: Yeah, we could factor out the non-lua logic in the future, one thing I can think of is potentially having features that allow virtual rooms to get loaded automatically, based on xml for instance. That may seem counterintuitive for virtual rooms as a whole, but I could see it being an option for boss related stuff due to the ID issues we've discussed before.

repentogon/Patches/VirtualRoomSets.cpp
namishere: ~~scatmans-world branch has a patch for RoomConfig_Room destructor that Sylmir made to fix heap corruption that was occurring. you can find the patch via his thoughts on it, "Nightmare stuff."~~
it's for RoomSet, not RoomConfig_Room. maybe still worth looking at? https://github.com/TeamREPENTOGON/REPENTOGON/commit/a1858eaa4981193465904f48a6cf6079b6503135

Sylmir: The patch for RoomSet was due to the fact that we manually instanciate RoomSet on Repentogon's side, meaning the compiler inserts a call to our destructor (which implicitly calls the destructor of every subobject, recursively). However, our destructor used to call the game destructor as well. This meant that every subobject's destructor was called twice. The solution was to patch our own destructor to immediately jump into the game's destructor, so that objects are destroyed only once.

If something you're doing here can lead to a call to a destructor of a base game object, we may run into heap corruptions down the line. It all depends on how trivial the structures you are destroying are. If they are trivial (no memory management required) then it should be fine. If they contain managed resources, we're at risk.

PR #825, opened 2026-02-20
Image Lua API

Guantol-Lemat:
- Adds Lua methods to the `Renderer`, `Image` and `Quad` classes, primarily focused on enabling rendering using low-level images and custom shaders.
- An internal ShaderLoader for custom shaders has been added. It's used by the newly added `Renderer.LoadShader` method, and it replaces the old loader for ANM2 Custom Shaders.
- Additional checks have been added for ANM2 Custom Shaders, to avoid usage of invalid shaders during load or after a failed reload.

Added Lua Methods:
- Renderer
    - CreateImage()
    - RenderToSurface()
    - GetShaderByType()
    - GetPixelationAmount()
    - GetClipPaneNormal()
    - GetClipPaneThreshold()
    - LoadShader()
- Image
	- GetWidth()
	- GetHeight()
	- GetPaddedWidth()
	- GetPaddedHeight()
	- GetName()
	- GetTexelRegion()
	- Render()
	- RenderWithShader()
- DestinationQuad
	- Copy()
	- Translate()
	- Scale()
	- Rotate()
	- Shear()
	- ApplyMatrix()
	- Flip()
	- __tostring
- SourceQuad
	- Copy()
	- IsUVSpace()
	- ConvertToPixelSpace()
	- ConvertToUVSpace()
	- __tostring

### DIFF
--- a/repentogon/LuaInterfaces/LuaRender.cpp
+++ b/repentogon/LuaInterfaces/LuaRender.cpp
@@ -2769,6 +2769,12 @@ static void RegisterCustomRenderMetatables(lua_State* L) {
 
 LUA_FUNCTION(lua_Renderer_LoadImage) {
 	const char* path = luaL_checkstring(L, 1);
+	std::filesystem::path p = path;
+	if (!p.is_relative())
+	{
+		return luaL_error(L, "Image %s does not exist", path);
+	}
+	
 	KAGE_SmartPointer_ImageBase image;
 	Manager::LoadImage(&image, path, __ptr_g_VertexAttributeDescriptor_Position, false);
 


repentogon/LuaInterfaces/LuaRender.cpp
```diff
@@ -2290,6 +2783,69 @@ LUA_FUNCTION(lua_Renderer_LoadImage) {
 	return 1;
 }
 
+LUA_FUNCTION(Lua_Renderer_CreateImage) {
+	uint32_t width = (uint32_t)luaL_checkinteger(L, 1);
+	uint32_t height = (uint32_t)luaL_checkinteger(L, 2);
+	const char* name = luaL_checkstring(L, 3);
```
ConnorForan, line 2795: So each procedural image created by mods are required to have a completely unique name string?

I don't see this as an issue, just confirming. Also good that we prevent clashes with real images at least.

repentogon/LuaInterfaces/LuaRender.cpp
```diff
@@ -2290,6 +2783,69 @@ LUA_FUNCTION(lua_Renderer_LoadImage) {
...
+	std::string trueName = REPENTOGON::StringFormat("%s.procedural", name);
+	KAGE_SmartPointer_ImageBase pointer = KAGE_Graphics_ImageManager::CreateProceduralImage(width, height, trueName.c_str(), color);
+
+	if (!pointer.image) {
+		return luaL_error(L, "Unable to create Image");
+	}
+
+	LuaImage* ud = new (lua_newuserdata(L, sizeof(LuaImage))) LuaImage;
+	memset(ud, 0, sizeof(LuaImage));
+	luaL_setmetatable(L, LuaRender::ImageMT);
```
ConnorForan, line 2808: Are Images owned by the lua userdata, and properly/completely destroyed and cleaned up internally if garbage collected luaside?

repentogon/LuaInterfaces/LuaRender.cpp
```diff
@@ -2290,6 +2783,69 @@ LUA_FUNCTION(lua_Renderer_LoadImage) {
...
 }
 
+LUA_FUNCTION(Lua_Renderer_CreateImage) {
+	uint32_t width = (uint32_t)luaL_checkinteger(L, 1);
+	uint32_t height = (uint32_t)luaL_checkinteger(L, 2);
+	const char* name = luaL_checkstring(L, 3);
+	KColor color = KColor(0.0, 0.0, 0.0, 0.0);
+
+	// prevent name clashes with real images (since these are added to the cache for some reason, even though they are never loaded)
+	std::string trueName = REPENTOGON::StringFormat("%s.procedural", name);
```
ConnorForan, line 2799: Mostly just curious, are the "real images" here actual png files or just procedural images?

If there are pngs, is their "name" the filename? For vanilla procedural images, what sort of names do they use?

repentogon/LuaInterfaces/LuaRender.cpp
```diff
@@ -2290,6 +2783,69 @@ LUA_FUNCTION(lua_Renderer_LoadImage) {
...
+	}
+
+	LuaImage* ud = new (lua_newuserdata(L, sizeof(LuaImage))) LuaImage;
+	memset(ud, 0, sizeof(LuaImage));
+	luaL_setmetatable(L, LuaRender::ImageMT);
+	ud->image = pointer;
+	return 1;
+}
+
+LUA_FUNCTION(Lua_Renderer_RenderToSurface) {
```
ConnorForan: Should this be called RenderToImage? (Actually just asking)

For our lua API purposes it seems like the target is always known as an Image

repentogon/LuaInterfaces/LuaRender.cpp
Guantol-Lemat: A unique name is not really necessary. Unlike regular images the cache is not used during the creation of Procedural Images, even though they are added to the cache.

repentogon/LuaInterfaces/LuaRender.cpp
Guantol-Lemat: Yes, I checked and everything is properly freed and cleaned up.

repentogon/LuaInterfaces/LuaRender.cpp
Guantol-Lemat: Yes, I'm referring to actual files, which will have their name set to the filePath.

The reason `.procedural` is being added is simply because procedural images are, for some reason, added to the cache. If someone were to name an image like an actual filePath, the cache will start picking the procedural image instead of loading the actual file.

Vanilla procedural images have pretty standard name like "Floor Surface".

repentogon/LuaInterfaces/LuaRender.cpp
Guantol-Lemat: I've mostly seen render targets either being defined as RenderTexture or Surface in other APIs.

In our case not every image is a valid render target, so maybe we should change CreateImage into CreateSurface?

ConnorForan: Due to the Image userdata class I think it's fine to keep CreateImage named as-is. The distinction doesn't seem to be super important. I think Image is generally an intuitive name for that class.

So per our lua APIs you will be rendering to Images, but if you're content with the current naming of RenderToSurface its not a huge deal.

PR #855, opened 2026-05-05
improve linux launch command sed script

torzod:
this small change improves the sed command used in steam's launch options on linux in 2 ways:
* removes the unnecessary -e, as it is only needed if multiple scripts are to be run in a single sed command
* changes the delimiter in the sed script to | from / as doing so removes the requirement for the contained / to be escaped. this additionally makes it clearer to read since the script includes file paths that contain / as the separator.


Sylmir:
Thank you !

PR #879, opened 2026-07-09
Pathfinding FindGridPath CheckLine Grid Path Override

Syntax-Sculptor:
this is the best PR name to grace REPENTOGON.


repentogon/Patches/EntityPlus.h
```diff
@@ -11,6 +11,7 @@ class EntityPlus {
   public:
 	virtual ~EntityPlus() {}
 
+	int lineOfSightCostThreshold = 900;
```
ConnorForan: Is there a benefit to this being a threshold instead of a boolean setting? Like in practice, would you ever bother to set this to like 500 instead of just 0? Does it do anything to set this >900?

Legitimate question, if you can think of good justifications for these. The main benefit seems to be just ignoring everything <= 900

repentogon/LuaInterfaces/Entities/LuaEntity.cpp
```diff
@@ -861,6 +861,42 @@ LUA_FUNCTION(Lua_EntitySetVariant) {
...
+		lua_pushinteger(L, plus->lineOfSightCostThreshold);
+	}
+	else {
+		lua_pushinteger(L, 900);
+	}
+
+	return 1;
+}
+
+LUA_FUNCTION(Lua_EntitySetLineOfSightCostThreshold) {
```
ConnorForan: Had a thought, pretty sure that the Pathfinder class has a reference to its parent Entity: https://github.com/TeamREPENTOGON/REPENTOGON/blob/2bcd2358c109123a5a69383ce500b50a12ab116f/libzhl/functions/NPCAI_Pathfinder.zhl#L32

So in theory, you could still have these as Pathfinder functions, while under the hood they still read/write from EntityPlus

This is hypothetical though, lemme know if it doesn't seem like that would work. Just seems like something that should be in Pathfinder API-wise

Syntax-Sculptor: Good idea.

repentogon/Patches/EntityPlus.h
Syntax-Sculptor: I feel like there might be some niche cases but in retrospect, I think a boolean toggle will be more intuitive.

Syntax-Sculptor: In retrospect, since there's fireplaces and grid entities with their own grid paths being set, what if we opted for an enum instead? I'm thinking along the lines of 

```lua
GridPathThreshold = {
    DEFAULT = 900,
    IGNORE_DECAY = 950,
    IGNORE_FIREPLACES_AND_DECAY = 1000,
    IGNORE_EVERYTHING = 1001
}
```

I think these are pretty realistic situations that'd warrant an integer threshold over a boolean.

ConnorForan: No, if we keep the threshold route the "enum" is still just PathMarker/GridPath by another name (disregarding that the PathMarker enum doesn't exist in lua still, but yeah).

As a threshold this feature is just replacing the PathMarker/GridPath passed along to the CheckLine call inside FindGridPath - it cant be made more granular than that.

I suppose to actually consider the usefulness of the threshold, yeah you can increase it further to have something ignore fireplaces too, or other things if you go even higher. It's somewhat limited sure, but anything more granular will be much more complicated to implement, especially since part of the logic is baked into this CheckLine call.

You can keep the feature as a threshold, for that little extra potential usage. After looking at the decomp closer, turning this into a boolean in its current state might not be appropriate.

Though now that I am looking at it, shouldn't there be a second place where this threshold matters? Like, the current patch is for the direct path check, but there's a separate check done for the actual grid-based pathfinding that should also be handled, right?

PR #922, opened 2026-08-30
Fix modded filemap entries not properly generating

namishere:
Our custom filemap gen has had an issue since its creation where, seemingly arbitrarily, modded files would fail to load. Reports seemed to suggest it happening more often on Linux, but I don't have any evidence for that.

After lots of logging i found that in cases where the entry generation failed, trying to log `relpath` would immediately terminate the logged string.

Expected:
```
[2026-08-29 19:25:54] [FindFiles] I'm trying to add Z:/var/home/nami/.local/share/Steam/steamapps/common/The Binding of Isaac Rebirth/mods/the binding of isaac sacrilege soundtrack_3727170471\resources\music\end times.ogg with relpath music\end times.ogg!
[2026-08-29 19:25:54] [AddModEntry] I'm adding music/end times.ogg under hash 1412984731!

```
Failure:
```
[2026-08-29 19:25:54] [FindFiles] I'm trying to add Z:/var/home/nami/.local/share/Steam/steamapps/common/The Binding of Isaac Rebirth/mods/the binding of isaac sacrilege soundtrack_3727170471\resources\music\fight ogg\be done intro.ogg with relpath [2026-08-29 19:25:54] [AddModEntry] I'm adding
```

I still don't understand the root issue, but changing `relpath` from `std::wstring_view` to a standard `std::wstring` has fixed it.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,4 @@
-v1.1.2e -
+v1.1.2h -
 Added:
 * Isaac:
 	- OpenConsole()
@@ -40,6 +40,7 @@ Fixes:
 	- Fixed being unable to add modded items to itempools.xml if they use a localization string name.
 	- Fixed specifying a tainted counterpart using bskinparent in players.xml not working properly for localized names.
 	- Content .anm2 file animations that use the character's name will now fall back to the English name (or raw string) if an animation matching the current localized name is not found.
+* Fixed modded filemap generation setting sometimes failing to load files
 /newline/
 /versionseparator/
 v1.1.2g -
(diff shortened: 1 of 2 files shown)


namishere:
Replicatable with https://steamcommunity.com/sharedfiles/filedetails/?id=3727170471 via `MusicManager():Play(Music.MUSIC_BOSS2)`.

namishere:
Discussion with Sylmir shows that the issue is improper use of `wstring_view` on a temporary object `file.path()` causing `relpath` to point to a destroyed string. `wstring` creates a new string unbound to anything passed into its constructor. Could use `wstring&&` to extend `file.path()` lifetime, but for simplicity's sake this solution will work.

PR #927, opened 2026-09-06
Replace old fandom links in EntityPlayer.md

AllMeatball:
Fandom is not reliable anymore due to the amount of griefing Fandom permits on their platform, and being full of bloat (intrusive ads, superflous sidebar, etc). it shouldn't be used for documentation links. wiki.gg is a better vetted source now  and is official(?).

This PR is related to [the one for IsaacDocs](https://github.com/wofsauge/IsaacDocs/pull/538).


jerbmega:
Fantastic! I completely forgot we had those.

PR #929, opened 2026-09-06
Online co-op with mods: determinism measures and the mod-ban question

DiegoArias32:
What this is
Work on making online co-op reachable with mods loaded, plus four measures aimed at the desync that follows once you get there. Verified live to the point where a run starts with 129 mods loaded; not yet verified with two players.

Opening this as a discussion piece rather than a merge-ready change - see "What needs deciding" below, since one part of it deliberately reverses an existing decision.

The determinism measures
These stand on their own regardless of what happens with the online part.

Deterministic mod ordering. ModManager::ListMods fills _mods in filesystem enumeration order, which is a property of the disk rather than of the mod set - install order, drive, filesystem. Modded content IDs are assigned by walking that vector, so the same item can hold a different ID on two machines that have identical mods. Now sorted on workshop ID, or directory name for local mods.

Run-seeded Lua RNG. math.random is seeded per process, so identical mod code on an identical run seed still draws different numbers - a guaranteed desync for any mod that reaches for it. Deliberately not routed into the game's RNG: that stream is game state and consuming from it would change every other roll. A separate generator seeded from the run seed keeps machines in step while leaving the game's stream untouched.

This also fixes a real defect found while writing it: Lua 5.3 gives math.random(0) its own meaning - an integer with every bit random, not a range. Worth having independently of anything else here.

Sorted pairs(). Lua's iteration order depends on table layout and pointer values, so a mod taking the first match out of a pairs() loop diverges with nothing visibly wrong in its code. Installed but off by default: it changes behaviour for every pairs() call in every mod and in REPENTOGON's own Lua, and a second machine running these mods died with a stack buffer overrun that has not been explained yet. SetDeterministicPairs(true) enables it.

Unlock fingerprint. Measurement only. Two players have different save files, and the item pool is filtered by unlock state, so a treasure room rolls against different contents from the same seed. Unlock queries are folded into a rolling fingerprint so two players can compare one number instead of guessing. Deliberately does not change anything - see below.

What needs deciding
The _modBanStatus change reverses a deliberate decision. The hook on ModManager::ListMods currently forces 3 with the comment "prevents playing online modes". This branch forces 0 instead, which is a deliberate lie to the game's own check: the mods really are loaded. That is the whole reason online becomes reachable, and it is not my call to make for the project.

For what it is worth, reading the writers in the binary: initialisation clears the field to 0 and sets 3 if the global at 0xBE73C4 is above zero; the per-mod loop sets 2 for one class of mod, 1 for another, and escalates to 4 when both are present. 3 is terminal. So it tracks which mods are loaded rather than how bad they are.

Nothing here syncs unlock state between players, which is the cause identical mod sets do not rule out. Forcing agreement means deciding whose unlocks win, which changes what a player can obtain in their own run - a product decision, not a technical one, which is why the fingerprint measures rather than fixes.

Not merge-ready as it stands
Being upfront: this branch also carries debugging instrumentation from the investigation ([ONLINE-TRACE], [HOOK-ORDER], per-frame menu logging) and an experiment in XMLData.cpp that narrows a menu guard. Those want stripping before any merge.

If there is interest in only the defensible parts - mod ordering, the math.random(0) fix, the seeded RNG - I am happy to open a separate clean branch with just those and none of the online bypass.


jerbmega:
> If there is interest in only the defensible parts - mod ordering, the math.random(0) fix, the seeded RNG - I am happy to open a separate clean branch with just those and none of the online bypass.

There is no interest in any of this, really. Not only are we not going to accept vibe coded slop into REPENTOGON, but you seem to not understand that introducing online support puts both us and mods on the line to fix any online-exclusive issues that are out of scope in a sane setup. 

Keep your experiments to yourself. If you want to contribute, don't regurgitate what God Emperor Claude has blindly fed you, and think about the long term implementation and support ramifications of adding a feature- especially a vibe coded one that we then have to troubleshoot and support indefinitely. REPENTOGON is a passion project that grew from people learning naturally. Most of us barely understood assembly when we started. We have better things to do that actually benefit mods, and we have no desire to take our limited time for this hobby and use it to wade through a vibe coded nightmare that doesn't actually understand what it's trying to implement in the first place.

